PR #223, opened 2026-04-11
feat: add applyMatrix method to Mobject (#217)

### LINKED ISSUE #217: add `applyMatrix` method
## Description
In `manim`, the base `MObject` has a `apply_matrix` function that can transform points in a generic way. Currently, manim-web can animate using a matrix, but not apply the matrix directly.

## Use Case
It would allow more flexible object positioning and transformation.

## Proposed Solution
Implement the equivalent of:

```
def apply_matrix(  
    self,  
    matrix: MatrixMN,  
    *,  
    about_point: Point3DLike | None = None,  
    about_edge: Vector3DLike | None = None,  
) -> Self:
```

Note that `about_point` is the origin of the transform, and `about_edge` indicated the direction one should look at to find the extremal point. For a square, `about_edge=RIGHT` means that we take the center of the right segment. `about_edge=RIGHT+UP` takes the top right corner.

Also not that `about_point` and `about_edge` should be supported for a lot of other transforms like `scale`

maloyan:
## Summary
- Add `applyMatrix(matrix, options?)` method to `Mobject` for instant matrix transformations (3x3 and 4x4)
- Support `aboutPoint` and `aboutEdge` options matching Manim Python's API
- Extract shared `transformPointByMatrix` utility used by both the new method and the existing `ApplyMatrix` animation
- Add example showing animated shear transformation

Closes #217

## Test plan
- [x] 8 unit tests covering identity, scaling, rotation, aboutPoint, aboutEdge, VGroup, 4x4, and chaining
- [x] All 5713 existing tests pass
- [x] Example works at `examples/apply_matrix_method.html`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rambip:
Wonderful !
Note that manim only supports 2x2 and 3x3 matrices, even for 3d objects (the 2x2 matrix is completed with a 1 to leave the z coordinate as-si). If you implemented the 4x4 matrix as a "translation + transform" matrix, it could lead to a confusing situation for users.
The "aboutPoint" argument already allows the user to set the "translation" part.

PR #230, opened 2026-04-11
Fix mathjax import for headless mode

rambip:
## Summary
"Headless mode" added support for mathjax without the browser: used from node.
Unfortunately, it was not added as a dependency and the path was wrong.

## Changes
- dep in package.json
- MathJaxRender import path

## Testing
- [x] Tests pass (`npm test`)

## Note

This will not change the bundle, just allow headless mode to work with mathjax. I did not add a test for it but that could be a good thing.


rambip:
I simplified the import logic: now we import mathjax-full diretly but it's in rollup external deps.

PR #253, opened 2026-04-13
Font size constant

rambip:
## Summary

Use same strategy to compute size of Tex as manim python.

## Changes
- added a constant for screen size -> font size

## Testing
- [x] Tests pass (`npm test`)

## Remaining work
- check visually that new version makes sense
- do same for Text if necessary


rambip:
(probably not ready to merge, I need help on this one)

rambip:
I was so confused by the font-size calculation in manim that I opened an issue on their repo:
https://github.com/ManimCommunity/manim/issues/4690

rambip:
@maloyan Done :heavy_check_mark: 

I need your review. Can you check that the examples render correctly ? The font size is not exactly the same as before but it should match manim.

src/constants/fontRender.ts
```diff
@@ -0,0 +1,38 @@
...
+const LATEX_DEFAULT_FONT_SIZE_PT = 10;
+export const DEFAULT_FONT_SIZE_PT = 48;
+export const KATEX_REFERENCE_FONT_SIZE_PX = LATEX_DEFAULT_FONT_SIZE_PT * PT_TO_PX;
+export const MATHJAX_SVG_UNITS_PER_EM = 1000;
+
+/**
+ * The convention taken by Manim:
+ * How many world units a font of DEFAULT_FONT_SIZE_PT should occupy.
+ */
+export const DEFAULT_FONTSIZE_TO_WORLD_SPACE = 0.5 * PT_TO_PX;
```
maloyan: This value produces `DEFAULT_FONTSIZE_TO_WORLD_SPACE = 0.5 * 4/3 ≈ 0.667`, but the legacy scaling convention was `PIXEL_TO_WORLD = 0.01` (100 pixels = 1 world unit). At the default 48pt, the old code gave `48 * 0.01 = 0.48` world units — this gives `0.667`, making all `Text` ~39% bigger and `MathTexImage` ~5× bigger.

Fix: replace with `DEFAULT_FONT_SIZE_PT * 0.01 = 0.48` to preserve visual parity.

rambip:
>  but the legacy scaling convention was PIXEL_TO_WORLD = 0.01

Did the old convention match manim previously ?
Do you know if a MathTex $x$ was the same size as a text "x" ? The current implementation is supposed to enforce this.

Do you think there is a way to add an option at the scene level so that user can chose between "legacy" and "match manim" ?

> MathTexImage ~5× bigger.

This is probably a bug, did you check it visually ? MathTexImage should exactly the same size as svg.

I was not able to check the examples, `docusaurus` takes up all my ram :neutral_face:

rambip:
Previous state:

maloyan:
Thanks for the detailed questions @rambip! I investigated the scaling math. Here's what I found:

### 1. "Did the old convention match manim previously?"

No — the old convention used a simple `PIXEL_TO_WORLD = 0.01` (100 pixels = 1 world unit) which was a pragmatic choice, not derived from Python Manim's constants. Your PR correctly tries to align with Manim's approach, but the specific value of `DEFAULT_FONTSIZE_TO_WORLD_SPACE = 0.5 * PT_TO_PX ≈ 0.667` causes a visible size increase compared to the old behavior.

### 2. "Do you know if MathTex `x` was the same size as Text `x`?"

In the old code, they were **not** exactly the same size — `Text` and `MathTexImage` both used `scaleFactor = 0.01` but had different rendering pipelines (canvas pixels vs MathJax SVG units). Your PR's goal of unifying them through shared constants is the right direction.

### 3. "MathTexImage should be exactly the same size as SVG"

I found **two separate scaling issues** in `MathTexImage`:

- **KaTeX path**: The container now renders at `KATEX_REFERENCE_FONT_SIZE_PX` (13.33px) instead of `this._fontSize` (e.g. 48px), but the world-scale conversion `fontSize * FONT_PT_TO_WORLD` doesn't fully compensate. Result: **~1.39× bigger** at 48pt.

- **MathJax path in MathTexImage** (not MathTex SVG): The old code used `fontScale = fontSize/48` to pre-scale the SVG dimensions before applying `scaleFactor = 0.01`. The PR removed `fontScale` but now applies `fontSize * FONT_PT_TO_WORLD = 0.05` instead of `0.01`. Result: **~5× bigger** at 48pt. This one is likely a bug — the fontScale removal wasn't compensated in the world-scale calculation.

### 4. "Is there a way to add a scene-level option?"

That could work as a future enhancement, but I'd suggest first getting the constants right so both "legacy" and "match manim" produce correct output, then expose the choice. The current PR has the right architecture (single source of truth in `fontRender.ts`) — just the baseline value needs adjustment.

### Summary

The `MathTex` (SVG) scaling looks correct with the new constants. The issues are in `Text` and `MathTexImage` where the old `0.01` convention isn't fully preserved. The `DEFAULT_FONTSIZE_TO_WORLD_SPACE` value is the root — `0.5 * PT_TO_PX = 0.667` vs the old effective value of `0.48`.

maloyan:
Also @rambip — you don't need to build the full Docusaurus site to test visually. You can run individual examples directly:

```bash
npx vite
```

Then open `http://localhost:5173/examples/mathtex_svg.html` (or any other `.html` file in `examples/`). Much lighter on RAM than a full docs build.

rambip:
I'm making progress:

Before:
image


After:
image

rambip:
I would say there is a ~1.3 factor between first line and second line. I think it matches manim now. I should check.
If it does match manim, are you satisfied with this result  ?Or do you want me to add an option at the scene level for backward compatibility ?

If it is the case I can clean up the code, commit and let you merge.

Also: I noticed that I'm never able to get all the MathTex when I'm rendering them at the same time:

image


Maybe you can investigate ?

maloyan:
I guess that's good. Could you commit so I can check that locally once more

rambip:
Almost there, but I found another bug: when strokeWidth is set to zero for MathTex, the glyph does not render. But when set to `0.01`, it does.
strokeWidth should be 0 by default, since the svg provides a "filled path" (the countour)

rambip:
almost there !
image

Just a few tests to correct and a scale factor to investigate for tex image.

rambip:
image
pixel-perfect :star_struck:

rambip:
https://github.com/user-attachments/assets/59c7620f-02fb-484d-82e0-fe5fa6c71da5

This was a deep dive. I tried to document everything I learned along the way. I hope it will be useful for stability and clarity in the future.

PR #279, opened 2026-04-21
Issue 251 mathtex animation tests

### LINKED ISSUE #251: Animations do not work for SVG-based `MathTex`
## Description
When doing any kind of transform from `MathTex`, nothing happens.

## Steps to Reproduce
```js
const tex = new MathTexSVG({ latex: 'x' });
scene.play(tex.animate.shift(RIGHT))
```
Nothing happens.

## Investigation

```js
const tex = new mathtexsvg({ latex: 'x'});
await tex.waitforrender();
tex.shift([6, 0, 0]);
scene.add(tex);  // position is correct
```

```js
const tex = new MathTexSVG({ latex: 'x' });
console.log('before', tex.children.length); // likely 0
console.log(tex.points[0]); // undefined
await tex.waitForRender();
console.log('after', tex.children.length); // >0
console.log(tex.points[1]); // defined
scene.play(tex.animate.shift(RIGHT)); // should move
```

So, I tried a fix in `Group.ts` and `VGoup.ts`

```js
 * @param delta - Translation vector [x, y, z]
 * @returns this for chaining
 */
override shift(delta: Vector3Tuple): this {
  if (this.children.length === 0) {
    super.shift(delta);
    return this;
  }
  for (const child of this.children) {
    child.shift(delta);
  }
  this._markDirty();
```

But this naive fix did not work.

90% sure it's related to the async rendering pipeline, but I don't know enough about it.

A few informations about it for those like me:
```
MathTexSVG Implementation

MathTexSVG uses a promise-based approach:

    Stores the render promise in _renderPromise MathTexSVG.ts:55-58
    Provides waitForRender() method to await completion MathTexSVG.ts:102-109
    Captures errors in _renderError for later inspection MathTexSVG.ts:187-190

Scene Integration

The Scene class automatically handles async rendering when mobjects are added:

    Calls _awaitAsyncRenders() for each added mobject Scene.ts:349-352
    Recursively checks children for async rendering Scene.ts:728-730
    Re-renders the scene when async rendering completes Scene.ts:716-722
    Logs warnings for failed renders Scene.ts:723-725
```

So we need a way to emit an animate command on an object that is not rendered yet, maybe by queuing or just applying the .animate on a later phase ?

---

EDIT: after a few more investigations, the problem might also come from:
- the fact that `tex.children[0].position` change but not `tex.position`
- the fact that  repeating `scene.play(tex.animate.shift())` only shifts once. (see #252 )

rambip:
## Summary
Fix `animate.shift()` for nested `MathTex`/`VGroup` hierarchies by matching and interpolating **leaf VMobjects** using effective (world) transforms, instead of relying on direct-child pairing.

## Changes
- Added shared traversal/transform helpers in `src/core/MobjectTraversal.ts`:
  - `collectLeafVMobjectSnapshots()`
  - `collectLeafVMobjects()`
  - `worldToParentLocalPosition()`
- Refactored `Transform._beginVGroup()` in `src/animation/transform/Transform.ts` to:
  - pair source/target by leaf snapshots,
  - compute leaf position interpolation in source parent-local space derived from world transforms,
  - correctly handle nested `VGroup` trees (e.g., multipart `MathTex`).
- Replaced broad/brittle MathTex animation tests with focused issue regression tests in `src/mobjects/text/mathtex-animation.test.ts`:
  - simple `MathTex.animate.shift`
  - multipart `MathTex` (nested VGroups)
  - repeated shift accumulation
  - VMobject-in-VGroup shift
  - mixed VGroup (MathTex + shape) shift
- Removed debug/broken test coverage not relevant to this fix.

## Testing
- [ ] Tests pass (`npm test`)
- [ ] Lint passes (`npm run lint`)

## Related Issues
Closes #251

**Caveat (current implementation):**
- Leaf matching is DFS-order based (structural), not semantic/id-based. If source/target reorder leaves, pairing may be suboptimal for morph quality.
- This PR is scoped to shift regression behavior; broader scale/rotate matching policy for complex nested groups remains future work.


rambip:
https://github.com/user-attachments/assets/167ed63a-6078-4c5a-9c7b-5343d85f014e

All the animations work. As you see, there is still a bit of a problem with the rendering of SVG path, but I think it's a separate issue.

rambip:
Also note that for the font size we will have to use 48 points because I already merged the previous PR.

maloyan (approved):

PR #282, opened 2026-04-22
feat: add autoCenter option to Swap and CyclicReplace

### LINKED ISSUE #281: Swap/CyclicReplace don't work with setPoints3D (position vs points mismatch)
## Description

manim-web has a `position` property that is separate from points, but animations like `ApplyPointwiseFunction` modify points directly. This causes animations like `Swap` and `CyclicReplace` to behave incorrectly.

The core issue: these animations move the `position` property, not the points themselves. When points have been modified but `position` is still `(0,0,0)`, the animations have no visible effect.

## Minimal Reproducible Example

```html
<!DOCTYPE html>
<html>
<head>
    <title>Swap MRE - manim-web position vs points</title>
</head>
<body>
    <h1>Swap doesn't work after ApplyPointwiseFunction</h1>
    <div id="container"></div>
    <script type="module">
        import {
            Scene,
            Circle,
            ApplyPointwiseFunction,
            Swap,
        } from "https://cdn.jsdelivr.net/npm/manim-web@0.3.18/dist/manim-web.browser.js";

        const scene = new Scene(document.getElementById("container"), { width: 500, height: 300 });

        // Create two circles at origin (position = 0,0,0)
        const c1 = new Circle({ radius: 0.4, color: "#FF0000" });
        const c2 = new Circle({ radius: 0.4, color: "#0000FF" });
        
        scene.add(c1, c2);
        
        // Shift circles using ApplyPointwiseFunction (modifies points, NOT position)
        // This simulates what Python manim's shift() does
        await scene.play(new ApplyPointwiseFunction(c1, (p) => [p[0] - 1, p[1], p[2]]));
        await scene.play(new ApplyPointwiseFunction(c2, (p) => [p[0] + 1, p[1], p[2]]));
        
        console.log("Before Swap:");
        console.log("c1.position:", c1.position.x, c1.position.y); // 0, 0
        console.log("c2.position:", c2.position.x, c2.position.y); // 0, 0
        console.log("c1 center:", c1.getCenter()); // -1, 0
        console.log("c2 center:", c2.getCenter()); // +1, 0
        
        // Swap should exchange positions, but position is (0,0,0) for both!
        // Expected: c1 moves to x=+1, c2 moves to x=-1
        // Actual: nothing moves!
        await scene.play(new Swap(c1, c2));
    </script>
</body>
</html>
```

## Expected Behavior

Swap should exchange the visual positions of the two objects. After the ApplyPointwiseFunction:
- c1 is visually at x=-1 (center)
- c2 is visually at x=+1 (center)
- Swap should move c1 to x=+1 and c2 to x=-1

## Actual Behavior

Objects don't move because Swap only swaps the `position` property, which is `(0,0,0)` for both circles.

Console output shows:
```
Before Swap:
c1.position: 0 0
c2.position: 0 0
c1 center: [-1, 0, 0]
c2 center: [1, 0, 0]
After Swap:
c1.position: 0 0  (should be 1 0)
c2.position: 0 0  (should be -1 0)
```

## Environment

- Browser: Any
- OS: Any
- manim-web version: 0.3.18

## Root Cause

- `ApplyPointwiseFunction` modifies points directly, not `position`
- `Swap` and `CyclicReplace` only move `position`, not points
- This mismatch causes these animations to fail when points have been modified without updating `position`

## Proposed Fix

Add a method like `centerPointsAroundPosition()` that:
1. Calculates the bounding box center of current points
2. Translates all points by `-center`
3. Sets `position = position + center`

Example:
```javascript
const c1 = new Circle();
c1.setPoints3D(pointsAtXMinus1); // World coords, position=(0,0,0)
c1.centerPointsAroundPosition(); // Now points centered at origin, position=(-1,0,0)

// Swap now works correctly!
await scene.play(new Swap(c1, c2));
```

## Additional Context

This is particularly important for:
1. Widget/integration use cases where mobject state is serialized from Python manim
2. Any workflow using `ApplyPointwiseFunction` followed by position-based animations

### Rotation Consideration

Rotation may have similar issues - rotating around center when position doesn't match point center will rotate around wrong axis.

rambip:
- Added centerPointsAroundPosition() method to VMobject
- Swap and CyclicReplace now auto-center by default (autoCenter: true)
- This fixes position/points mismatch after setPoints3D or ApplyPointwiseFunction
- Set autoCenter: false for old behavior (position-only)
- All tests restored and passing
 

## Changes
                                                                               
 New feature: centerPointsAroundPosition() on VMobject                                     
 - Added method to center points around origin and update position accordingly             
 - Fixes position/points mismatch when points are modified directly (via setPoints3D,      
 ApplyPointwiseFunction, etc.)                                                             
                                        
⚠️ **BREAKING CHANGE**:                                                   
 Swap and CyclicReplace auto-center by default, to match python manim.
- opt out with  `{autoCenter: false}`      
                                                                                           
 Tests                                                                                     
 - Restored old tests with autoCenter: false                                               
 - Added new tests verifying Swap/CyclicReplace work after ApplyPointwiseFunction     

## Testing
- [x] Tests pass (`npm test`)

## Related Issues
Closes #281


maloyan:
## Codex review summary

**High**: `centerPointsAroundPosition()` handles only `this._points3D` — child-backed geometry (`MathTex`, `SVGMobject`, `CurvesAsSubmobjects`, many `VGroup`s) is still broken, so `autoCenter` silently no-ops for the most common containers this PR is meant to help.
**Medium**: `(m as VMobject).centerPointsAroundPosition?.()` silently no-ops for non-VMobject inputs and hides a real type mismatch. Use a type guard / narrower signature.
**Low**: default flip to `autoCenter: true` is a breaking change with no `CHANGELOG.md` / release-note update.
**Low**: `BRANCH_STATUS.md` is branch-bookkeeping scratch — please do not commit `.md` files like this; drop from PR.
**Low**: new MRE tests cover only leaf `sq()`; add a regression for a child-backed mobject (`MathTex` / `SVGMobject` / `VGroup` / `CurvesAsSubmobjects`).

**Positive**: preserving old assertions behind `autoCenter: false` keeps non-centered path coverage; leaf MREs capture the original square regression clearly.

Inline comments left on each site. _(Review via codex.)_

src/core/VMobject.ts
```diff
@@ -572,6 +572,63 @@ export class VMobject extends VMobjectRendering {
...
+   * - Points are centered at the origin (0,0,0)
+   * - Position is updated to reflect the visual center
+   *
+   * This enables position-based animations like Swap and CyclicReplace to work
+   * correctly after point modifications.
+   *
+   * @returns this for chaining
+   */
+  centerPointsAroundPosition(): this {
+    if (this._points3D.length === 0) {
```
maloyan: **High — correctness gap.** Early-returning on `_points3D.length === 0` means this helper never fixes child-backed VMobjects/containers (`MathTex`, `SVGMobject`, `CurvesAsSubmobjects`, many `VGroup`s). The bbox scan below also duplicates `getCenter()` (line 548) but only for local points, so `autoCenter` still misses the family-backed cases this PR is trying to unblock. Re-center descendant geometry too, or extract shared child-aware center logic and translate the whole family.

src/animation/transform/MovementTransforms.ts
```diff
@@ -283,11 +288,17 @@ export class Swap extends Animation {
...
     this.mobject2 = mobject2;
     this.pathArc = options.pathArc ?? Math.PI / 2;
+    this.autoCenter = options.autoCenter ?? true;
   }
 
   override begin(): void {
     super.begin();
 
+    if (this.autoCenter) {
+      (this.mobject as VMobject).centerPointsAroundPosition?.();
```
maloyan: **Medium — silent no-op hides type errors.** `Swap` still accepts generic `Mobject`, but this cast + optional call silently does nothing for unsupported inputs instead of making the incompatibility explicit. That hides real type errors and makes `autoCenter: true` look universal when it is not. Prefer a type guard / throw here, or narrow the API to mobjects that actually implement the helper. Same issue appears in `CyclicReplace.begin()` below.

BRANCH_STATUS.md
```diff
@@ -0,0 +1,35 @@
+# Branch Status Report
```
maloyan: **Do not commit `.md` scratch files.** This reads like private branch-tracking notes, not repository content. It does not affect runtime/tests/docs and will just add noise to the tree and package. Drop it from the PR and keep this context in the PR description or an issue. General rule: no `.md` files in commits unless they are real project docs (README, CHANGELOG, AGENTS).

src/animation/animation-transform-extra.test.ts
```diff
@@ -1022,24 +1028,49 @@ describe('Swap', () => {
...
 
     expect(m1.position.x).toBeCloseTo(0, 5);
     expect(m2.position.x).toBeCloseTo(8, 5);
   });
 
-  it('factory creates Swap', () => {
-    const anim = swap(sq(), sq(), { pathArc: 1 });
-    expect(anim).toBeInstanceOf(Swap);
-    expect(anim.pathArc).toBeCloseTo(1, 5);
+  it('swaps positions after ApplyPointwiseFunction modifies points', () => {
```
maloyan, line 1046: **Low — test gap.** Useful MRE, but it only exercises a leaf `sq()` VMobject. The risky gap is child-backed geometry, so add at least one regression using `MathTex`, `SVGMobject`, `CurvesAsSubmobjects`, or a `VGroup` whose children were modified by `ApplyPointwiseFunction`; the bug flagged on `centerPointsAroundPosition()` would slip through these tests as written.

rambip:
Fixed

PR #283, opened 2026-04-23
fix y and z being swapped for 3d axis

### LINKED ISSUE #256: scene and ThreeDAxes y and z directions are swapped
## Description
A clear description of the bug.

## Steps to Reproduce
1. Create any asymmetric object without using `coordsToPoint()`
2. See it's not oriented properly

## Expected Behavior
`scene` and `ThreeDAxes` to have all basis vectors pointing in the same direction

## Actual Behavior
The xy plane is vertical with respect to the `scene` coordinates and horizontal with respect to `ThreeDAxes` coordinates

## Environment
Browser: Brave 1.89.132
OS: Windows 11 10.0.26200
manim-web version: 0.3.19

## Additional Context
This scene showcases this with two sets of basis vectors, one for `scene` and one for the `ThreeDAxes`

```<!DOCTYPE html>
<html>
<head>
  <style>
    * { margin: 0; padding: 0; box-sizing: border-box; }
    body { background: #191919; display: flex; justify-content: center; align-items: center; width: 100vw; height: 100vh; overflow: hidden; }
    #container { width: 100vw; height: 100vh; }
  </style>
</head>
<body>
  <div id="container"></div>

  <script type="module">
    import {ThreeDAxes, Arrow3D, ThreeDScene, RED, RED_E, GREEN, GREEN_E, BLUE, BLUE_E} from 'https://esm.sh/manim-web';

    const scene = new ThreeDScene(document.getElementById('container'), {
      width: window.innerWidth,
      height: window.innerHeight,
      backgroundColor: '#191919',
      phi: 75 * (Math.PI / 180),
      theta: -45 * (Math.PI / 180),
      distance: 20,
      fov: 30,
      enableOrbitControls: true,
    });

    const axes = new ThreeDAxes({
      xRange: [-5, 5, 1],
      yRange: [-5, 5, 1],
      zRange: [-5, 5, 1],
      xLength: 10,
      yLength: 10,
      zLength: 10,
      showLabels: true,
    });

    const axesX = new Arrow3D({start: [0, 0, 0], end: [1, 0, 0], color: GREEN});
    const axesY = new Arrow3D({start: [0, 0, 0], end: [0, 1, 0], color: RED});
    const axesZ = new Arrow3D({start: [0, 0, 0], end: [0, 0, 1], color: BLUE});

    const sceneX = new Arrow3D({start: axes.coordsToPoint(1, 0, 0), end: axes.coordsToPoint(2, 0, 0), color: GREEN_E});
    const sceneY = new Arrow3D({start: axes.coordsToPoint(0, 0, 0), end: axes.coordsToPoint(0, 1, 0), color: RED_E});
    const sceneZ = new Arrow3D({start: axes.coordsToPoint(0, 0, 1), end: axes.coordsToPoint(0, 0, 2), color: BLUE_E});

    scene.add(axes, axesX, axesY, axesZ, sceneX, sceneY, sceneZ);

    await scene.wait(Infinity);
  </script>
</body>
</html>```

### PR DESCRIPTION
**Pull request by rambip**
## Summary
Fix 3D axes alignment and orbit rotation axis selection.

## Breaking Changes
- Removed `_m2t()` coordinate conversion (Manim → Three.js). The coordinate convention has changed: Manim X right/Y forward/Z up now maps directly to Three.js without transformation.
- `orbitControlsUp` now sets the rotation axis only at initialization (Three.js default behavior). The axis is captured by OrbitControls internally and cannot be changed after construction.

### Before
3D axes were misaligned - Y and Z were swapped:
image

### After
Axes are now correctly aligned:
https://github.com/user-attachments/assets/d95c6a9a-fb3b-4ffa-ae66-95e9d65b4451

Users can now choose the rotation axis:
- `"x"` - rotate around X axis
- `"y"` - rotate around Y axis  
- `"z"` - rotate around Z axis
- `"camera"` - use camera initial up (Three.js default)
image

## Technical Details
- Fixed gimbal lock in `getOrbitAngles()` at poles (phi ≈ 0 or π) by storing last theta value
- Three.js OrbitControls captures camera.up into internal quaternion at construction - cannot be changed after
- `orbitControlsUp` is applied once at OrbitControls construction

## Testing
- [x] Tests pass (npm test)
- [x] Lint passes (npm run lint)
- [x] Types check (npm run typecheck)

## Related Issues
Closes #256

### CODE REVIEW

**Comment by rambip** [comment]
Still have to check if camera angles match python

**Comment by rambip** [comment]
@copilot 

I want manim-web to match python in terms of cam axes and default view, but keep the same ability to orbit-control. The main axis of orbit-control should be z (around with phi, closer with theta), (theta=0 by default)

Change tests expectations if necessary.

Python docs:
====================

The 3D camera system in Manim consists of two main implementations: `ThreeDCamera` for Cairo rendering and `OpenGLCamera` for OpenGL rendering, both used within `ThreeDScene` classes.

## 3D Camera Parameters and Default Values

### ThreeDCamera (Cairo Renderer)
The `ThreeDCamera` class accepts these parameters with their default values [1](#0-0) :

| Parameter | Default Value | Description |
|-----------|---------------|-------------|
| `focal_distance` | 20.0 | Distance from camera to focal point |
| `shading_factor` | 0.2 | Intensity of 3D shading effects |
| `default_distance` | 5.0 | Default camera distance from origin |
| `light_source_start_point` | 9*DOWN + 7*LEFT + 10*OUT | Initial light source position |
| `should_apply_shading` | True | Whether to apply 3D shading |
| `exponential_projection` | False | Use exponential projection instead of linear |
| `phi` | 0 | Polar angle (radians) |
| `theta` | -90*DEGREES | Azimuthal angle (radians) |
| `gamma` | 0 | Camera roll angle (radians) |
| `zoom` | 1 | Zoom factor |

### OpenGLCamera (OpenGL Renderer)
The `OpenGLCamera` class has different defaults [2](#0-1) :

| Parameter | Default Value | Description |
|-----------|---------------|-------------|
| `focal_distance` | 2.0 | Distance from camera to focal point |
| `light_source_position` | [-10, 10, 10] | Light source position |
| `orthographic` | False | Use orthographic projection |
| `minimum_polar_angle` | -PI/2 | Minimum phi angle allowed |
| `maximum_polar_angle` | PI/2 | Maximum phi angle allowed |
| `euler_angles` | [0, 0, 0] | Initial camera orientation (theta, phi, gamma) |

## 3D Camera Logic

### Rotation and Orientation
Both cameras use spherical coordinates (phi, theta, gamma) for orientation:

- **Phi**: Polar angle from Z-axis [3](#0-2) 
- **Theta**: Azimuthal angle around Z-axis [4](#0-3) 
- **Gamma**: Camera roll around view direction [5](#0-4) 

The `ThreeDCamera` generates a rotation matrix by combining three rotations [6](#0-5) :
1. Rotation about Z-axis by (-theta - 90°)
2. Rotation about X-axis by (-phi)
3. Rotation about Z-axis by gamma

### Projection Logic
The `ThreeDCamera` projects 3D points to 2D using perspective projection [7](#0-6) :

```python
# Linear projection: factor = focal_distance / (focal_distance - z)
# Exponential projection: factor = exp(z / focal_distance)
points[:, i] *= factor * zoom  # Applied to x and y coordinates
```

### Shading Logic
3D shading is applied to VMobjects when `shade_in_3d` is True [8](#0-7) . The shading calculates lighting based on:
- Light source position
- Surface normals at start and end corners
- Shading factor intensity

## Scene Integration

### ThreeDScene Defaults
`ThreeDScene` provides default camera orientations [9](#0-8) :
- phi: 70°
- theta: -135°

`SpecialThreeDScene` overrides these with different defaults [10](#0-9) :
- phi: 70°  
- theta: -110°

### Camera Control Methods
`ThreeDScene` provides methods to control the camera:
- `set_camera_orientation()` - Instant camera positioning [11](#0-10) 
- `move_camera()` - Animated camera movement [12](#0-11) 
- `begin_ambient_camera_rotation()` - Start continuous rotation [13](#0-12) 

## Notes

The OpenGL and Cairo cameras have different parameter defaults and internal implementations, but both expose similar high-level interfaces through `ThreeDScene`. The OpenGL camera uses a model matrix approach while the Cairo camera uses explicit rotation matrices and projection formulas.

Wiki pages you might want to explore:
- [Core Architecture (ManimCommunity/manim)](/wiki/ManimCommunity/manim#2)

### Citations

**File:** manim/camera/three_d_camera.py (L40-53)
```python
    def __init__(
        self,
        focal_distance: float = 20.0,
        shading_factor: float = 0.2,
        default_distance: float = 5.0,
        light_source_start_point: Point3DLike = 9 * DOWN + 7 * LEFT + 10 * OUT,
        should_apply_shading: bool = True,
        exponential_projection: bool = False,
        phi: float = 0,
        theta: float = -90 * DEGREES,
        gamma: float = 0,
        zoom: float = 1,
        **kwargs: Any,
    ):
```

**File:** manim/camera/three_d_camera.py (L113-137)
```python
    def modified_rgbas(
        self, vmobject: VMobject, rgbas: FloatRGBA_Array
    ) -> FloatRGBA_Array:
        if not self.should_apply_shading:
            return rgbas
        if vmobject.shade_in_3d and (vmobject.get_num_points() > 0):
            light_source_point = self.light_source.points[0]
            if len(rgbas) < 2:
                shaded_rgbas = rgbas.repeat(2, axis=0)
            else:
                shaded_rgbas = np.array(rgbas[:2])
            shaded_rgbas[0, :3] = get_shaded_rgb(
                shaded_rgbas[0, :3],
                get_3d_vmob_start_corner(vmobject),
                get_3d_vmob_start_corner_unit_normal(vmobject),
                light_source_point,
            )
            shaded_rgbas[1, :3] = get_shaded_rgb(
                shaded_rgbas[1, :3],
                get_3d_vmob_end_corner(vmobject),
                get_3d_vmob_end_corner_unit_normal(vmobject),
                light_source_point,
            )
            return shaded_rgbas
        return rgbas
```

**File:** manim/camera/three_d_camera.py (L167-175)
```python
    def get_phi(self) -> float:
        """Returns the Polar angle (the angle off Z_AXIS) phi.

        Returns
        -------
        float
            The Polar angle in radians.
        """
        return self.phi_tracker.get_value()
```

**File:** manim/camera/three_d_camera.py (L177-185)
```python
    def get_theta(self) -> float:
        """Returns the Azimuthal i.e the angle that spins the camera around the Z_AXIS.

        Returns
        -------
        float
            The Azimuthal angle in radians.
        """
        return self.theta_tracker.get_value()
```

**File:** manim/camera/three_d_camera.py (L197-206)
```python
    def get_gamma(self) -> float:
        """Returns the rotation of the camera about the vector from the ORIGIN to the Camera.

        Returns
        -------
        float
            The angle of rotation of the camera about the vector
            from the ORIGIN to the Camera in radians
        """
        return self.gamma_tracker.get_value()
```

**File:** manim/camera/three_d_camera.py (L284-303)
```python
    def generate_rotation_matrix(self) -> MatrixMN:
        """Generates a rotation matrix based off the current position of the camera.

        Returns
        -------
        np.array
            The matrix corresponding to the current position of the camera.
        """
        phi = self.get_phi()
        theta = self.get_theta()
        gamma = self.get_gamma()
        matrices = [
            rotation_about_z(-theta - 90 * DEGREES),
            rotation_matrix(-phi, RIGHT),
            rotation_about_z(gamma),
        ]
        result = np.identity(3)
        for matrix in matrices:
            result = np.dot(matrix, result)
        return result
```

**File:** manim/camera/three_d_camera.py (L305-340)
```python
    def project_points(self, points: Point3D_Array) -> Point3D_Array:
        """Applies the current rotation_matrix as a projection
        matrix to the passed array of points.

        Parameters
        ----------
        points
            The list of points to project.

        Returns
        -------
        np.array
            The points after projecting.
        """
        frame_center = self.frame_center
        focal_distance = self.get_focal_distance()
        zoom = self.get_zoom()
        rot_matrix = self.get_rotation_matrix()

        points = points - frame_center
        points = np.dot(points, rot_matrix.T)
        zs = points[:, 2]
        for i in 0, 1:
            if self.exponential_projection:
                # Proper projection would involve multiplying
                # x and y by d / (d-z).  But for points with high
                # z value that causes weird artifacts, and applying
                # the exponential helps smooth it out.
                factor = np.exp(zs / focal_distance)
                lt0 = zs < 0
                factor[lt0] = focal_distance / (focal_distance - zs[lt0])
            else:
                factor = focal_distance / (focal_distance - zs)
                factor[(focal_distance - zs) < 0] = 10**6
            points[:, i] *= factor * zoom
        return points
```

**File:** manim/renderer/opengl_renderer.py (L95-108)
```python
    def __init__(
        self,
        frame_shape: tuple[float, float] | None = None,
        center_point: Point3DLike | None = None,
        # Theta, phi, gamma
        euler_angles: Point3DLike | None = None,
        focal_distance: float = 2.0,
        light_source_position: Point3DLike | None = None,
        orthographic: bool = False,
        minimum_polar_angle: float = -PI / 2,
        maximum_polar_angle: float = PI / 2,
        model_matrix: MatrixMN | None = None,
        **kwargs: Any,
    ) -> None:
```

**File:** manim/scene/three_d_scene.py (L46-49)
```python
            default_angled_camera_orientation_kwargs = {
                "phi": 70 * DEGREES,
                "theta": -135 * DEGREES,
            }
```

**File:** manim/scene/three_d_scene.py (L55-100)
```python
    def set_camera_orientation(
        self,
        phi: float | None = None,
        theta: float | None = None,
        gamma: float | None = None,
        zoom: float | None = None,
        focal_distance: float | None = None,
        frame_center: Mobject | Sequence[float] | None = None,
        **kwargs,
    ):
        """
        This method sets the orientation of the camera in the scene.

        Parameters
        ----------
        phi
            The polar angle i.e the angle between Z_AXIS and Camera through ORIGIN in radians.

        theta
            The azimuthal angle i.e the angle that spins the camera around the Z_AXIS.

        focal_distance
            The focal_distance of the Camera.

        gamma
            The rotation of the camera about the vector from the ORIGIN to the Camera.

        zoom
            The zoom factor of the scene.

        frame_center
            The new center of the camera frame in cartesian coordinates.

        """
        if phi is not None:
            self.renderer.camera.set_phi(phi)
        if theta is not None:
            self.renderer.camera.set_theta(theta)
        if focal_distance is not None:
            self.renderer.camera.set_focal_distance(focal_distance)
        if gamma is not None:
            self.renderer.camera.set_gamma(gamma)
        if zoom is not None:
            self.renderer.camera.set_zoom(zoom)
        if frame_center is not None:
            self.renderer.camera._frame_center.move_to(frame_center)
```

**File:** manim/scene/three_d_scene.py (L102-138)
```python
    def begin_ambient_camera_rotation(self, rate: float = 0.02, about: str = "theta"):
        """
        This method begins an ambient rotation of the camera about the Z_AXIS,
        in the anticlockwise direction

        Parameters
        ----------
        rate
            The rate at which the camera should rotate about the Z_AXIS.
            Negative rate means clockwise rotation.
        about
            one of 3 options: ["theta", "phi", "gamma"]. defaults to theta.
        """
        # TODO, use a ValueTracker for rate, so that it
        # can begin and end smoothly
        about: str = about.lower()
        try:
            if config.renderer == RendererType.CAIRO:
                trackers = {
                    "theta": self.camera.theta_tracker,
                    "phi": self.camera.phi_tracker,
                    "gamma": self.camera.gamma_tracker,
                }
                x: ValueTracker = trackers[about]
                x.add_updater(lambda m, dt: x.increment_value(rate * dt))
                self.add(x)
            elif config.renderer == RendererType.OPENGL:
                cam: OpenGLCamera = self.camera
                methods = {
                    "theta": cam.increment_theta,
                    "phi": cam.increment_phi,
                    "gamma": cam.increment_gamma,
                }
                cam.add_updater(lambda m, dt: methods[about](rate * dt))
                self.add(self.camera)
        except Exception as e:
            raise ValueError("Invalid ambient rotation angle.") from e
```

**File:** manim/scene/three_d_scene.py (L211-303)
```python
    def move_camera(
        self,
        phi: float | None = None,
        theta: float | None = None,
        gamma: float | None = None,
        zoom: float | None = None,
        focal_distance: float | None = None,
        frame_center: Mobject | Sequence[float] | None = None,
        added_anims: Iterable[Animation] = [],
        **kwargs,
    ):
        """
        This method animates the movement of the camera
        to the given spherical coordinates.

        Parameters
        ----------
        phi
            The polar angle i.e the angle between Z_AXIS and Camera through ORIGIN in radians.

        theta
            The azimuthal angle i.e the angle that spins the camera around the Z_AXIS.

        focal_distance
            The radial focal_distance between ORIGIN and Camera.

        gamma
            The rotation of the camera about the vector from the ORIGIN to the Camera.

        zoom
            The zoom factor of the camera.

        frame_center
            The new center of the camera frame in cartesian coordinates.

        added_anims
            Any other animations to be played at the same time.

        """
        anims = []

        if config.renderer == RendererType.CAIRO:
            self.camera: ThreeDCamera
            value_tracker_pairs = [
                (phi, self.camera.phi_tracker),
                (theta, self.camera.theta_tracker),
                (focal_distance, self.camera.focal_distance_tracker),
                (gamma, self.camera.gamma_tracker),
                (zoom, self.camera.zoom_tracker),
            ]
            for value, tracker in value_tracker_pairs:
                if value is not None:
                    anims.append(tracker.animate.set_value(value))
            if frame_center is not None:
                anims.append(self.camera._frame_center.animate.move_to(frame_center))
        elif config.renderer == RendererType.OPENGL:
            cam: OpenGLCamera = self.camera
            cam2 = cam.copy()
            methods = {
                "theta": cam2.set_theta,
                "phi": cam2.set_phi,
                "gamma": cam2.set_gamma,
                "zoom": cam2.scale,
                "frame_center": cam2.move_to,
            }
            if frame_center is not None:
                if isinstance(frame_center, OpenGLMobject):
                    frame_center = frame_center.get_center()
                frame_center = list(frame_center)

            zoom_value = None
            if zoom is not None:
                zoom_value = config.frame_height / (zoom * cam.height)

            for value, method in [
                [theta, "theta"],
                [phi, "phi"],
                [gamma, "gamma"],
                [zoom_value, "zoom"],
                [frame_center, "frame_center"],
            ]:
                if value is not None:
                    methods[method](value)

            if focal_distance is not None:
                warnings.warn(
                    "focal distance of OpenGLCamera can not be adjusted.",
                    stacklevel=2,
                )

            anims += [Transform(cam, cam2)]

        self.play(*anims + added_anims, **kwargs)
```

**File:** manim/scene/three_d_scene.py (L464-467)
```python
        default_angled_camera_position={
            "phi": 70 * DEGREES,
            "theta": -110 * DEGREES,
        },
```
==================================

rambip:
I ended up simplifying camera handling and using the same convention as manim.
This means you can easily have the camera identity matrix: this way you see your 3d scene as a 2d scene

image

rambip:
I reproduced the remaining orbit-controls issue on `fix-3d-axes` (using `examples/3d-scenes/three_d_axes_coords.html`).

What I tried:
- Set `orbitControlsUp: "x"` in the example and tested drag behavior.
- Added a wrapper fix in `src/interaction/OrbitControls.ts` to re-apply the chosen up-axis (`x/y/z`) on every `update()` (to counter `Camera3D.orbit()` rewriting `camera.up`).
- Also tried a second experiment that preserved screen-up after each update (projecting world +Y to the image plane).

What it showed:
- Re-applying up each frame keeps the orbit *constraint axis* stable, but does **not** fix the mouse interpretation issue.
- The screen-up preservation experiment makes interaction feel more Y-up, but that effectively defeats true X-up semantics (so it is not a correct fix).

Restore status:
- I reverted both experiments; branch is clean again.

Next theory:
- The remaining bug is likely not just "camera.up getting overwritten".
- More likely: there is a mismatch between our custom camera spherical convention (`Camera3D.orbit` / `theta,phi` frame) and OrbitControls' internal drag-to-azimuth/polar interpretation when non-Y up is used.
- So fix probably needs to happen at camera/control convention boundary (how initial basis / target / up are synced), not by post-correcting `camera.up` after the fact.

rambip:
Update from experiment results (thanks @rambip):

Data confirms stale OrbitControls basis at construction time:
- With `orbitControlsUp: "x"`, initial dumps showed `camera.up=[1,0,0]` but `storedQuatVsRefAngle=2.140638` and large `map(up)->Y` error.
- After programmatic camera move, mismatch persisted.
- After manually rebuilding `controls._quat/_quatInverse` from current `camera.up`, drag feel became natural (but axis semantics shifted due to current `camera.up` no longer being strict X in that state).

Interpretation:
- OrbitControls captures internal basis (`_quat`) from `camera.up` in its constructor.
- In our wrapper we were setting `camera.up` **after** `new ThreeOrbitControls(...)`, so controls had an inconsistent internal basis from frame 0.

Fix applied locally:
- Set `camera.up` from `orbitControlsUp` **before** constructing `ThreeOrbitControls`.
- Kept the rest of the wrapper behavior unchanged.

Next likely issue if any remains:
- `Camera3D.orbit()` also mutates `camera.up`, which can still desync semantics after programmatic camera changes. If needed, follow-up should decouple roll/up handling from orbit controls state (or explicitly re-sync basis when camera.up is intentionally changed).

maloyan:
## Codex review summary

**High** — `Camera3D.getOrbitAngles()` loses azimuth at the poles: at `phi=0` or `phi=π`, `atan2(0,0) → 0`, so any `getCameraOrientation()` → `setCameraOrientation()` roundtrip silently rewrites theta and roll. Preserve previous `theta` when `sin(phi) ≈ 0`.
**High** — `orbitControlsUp` is applied only at construction. Later `setCameraOrientation() / moveCamera() / orbit()` overwrite `camera.up`, while `ThreeOrbitControls` keeps the basis quaternion it captured. First programmatic move desyncs drag directions. Persist the chosen mode and reapply/rebuild the controls basis after every programmatic orbit.
**Medium** — Silent public-API break: `ThreeDAxes.coordsToPoint/pointToCoords/c2p{X,Y,Z}`, tick/label positions, and `Camera3D.orbit()` all now return different values for the same inputs vs the Manim→THREE mapping on `main`. No CHANGELOG/migration note in the PR. Add one or ship behind a compat switch/deprecation path; semver major.
**Low** — `orbit_controls_basis_experiment.html` mutates `controls._quat` / `_quatInverse` (private three.js internals) and reads like debug scratch. Either wrap in a supported API in our `OrbitControls` class or move it out of `examples/`.
**Low** — `three_d_axes_coords.html` mixes `phi: 0` (camera at +Z) with `orbitControlsUp: "x"` (screen-up = +X). Confusing as a regression demo for #256 — keep the default basis here, put x-up in a separate demo.

**Positive** — new `ThreeDAxes` regression tests are useful; `Camera3D.lookAtTarget` getter exists, so the preserved-target code path in `ThreeDScene` is valid. Removing the `_m2t` mapping does simplify the scene convention.

Inline comments left on each site. _(Review via codex.)_

src/core/Camera.ts
```diff
@@ -158,10 +171,13 @@ export class Camera3D {
...
 
-    // Inverse of orbit(): dx = d*sin(phi)*cos(theta), -dz = d*sin(phi)*sin(theta)
+    // Inverse of orbit():
+    // x = d*sin(phi)*cos(theta) -> cos(theta) = x/(d*sin(phi))
+    // y = d*sin(phi)*sin(theta) -> sin(theta) = y/(d*sin(phi))
+    // z = d*cos(phi)           -> cos(phi) = z/d
     return {
-      phi: Math.acos(dy / distance),
-      theta: Math.atan2(-dz, dx),
+      phi: Math.acos(dz / distance),
```
maloyan: **High — azimuth lost at the poles.** `orbit(0, -π/2, ...)` and `orbit(π, θ, ...)` both flatten `sin(phi) ≈ 0`, so `atan2(y, x)` here evaluates `atan2(0, 0) → 0`. Any `getCameraOrientation()` → `setCameraOrientation()` roundtrip then silently rewrites theta (and therefore roll) even though the camera did not move. Preserve the previous `theta` when `sin(phi) < ε`, or clamp/document the singularity explicitly.

src/core/ThreeDScene.ts
```diff
@@ -99,12 +102,23 @@ export class ThreeDScene extends Scene {
...
     if (enableOrbitControls && !this.isHeadless) {
-      this._orbitControls = new OrbitControls(this._camera3D.getCamera(), this.getCanvas(), {
+      const cam = this._camera3D.getCamera();
+      const preservedPos = cam.position.clone();
+      const preservedTarget = this._camera3D.lookAtTarget;
+
+      this._orbitControls = new OrbitControls(cam, this.getCanvas(), {
         enableDamping: true,
         dampingFactor: 0.05,
+        orbitControlsUp,
```
maloyan, line 112: **High — `orbitControlsUp` desyncs after first programmatic camera move.** The option is forwarded to `OrbitControls` at construction, which sets `camera.up` and then caches a basis quaternion. Every subsequent `setCameraOrientation() / moveCamera() / Camera3D.orbit()` overwrites `camera.up` from the gamma/phi/theta formula, but the cached basis in `ThreeOrbitControls` is not rebuilt — so drag directions stop matching the requested up axis. Persist the chosen mode on the scene and reapply / rebuild the controls basis after every programmatic orbit (or, better, drive up through a single authoritative path).

src/mobjects/three-d/ThreeDAxes.ts
```diff
@@ -374,23 +367,21 @@ export class ThreeDAxes extends Group {
   }
 
   /**
-   * Convert graph coordinates (Manim space) to visual point coordinates (THREE.js space).
+   * Convert graph coordinates to visual point coordinates (scene space).
    * With unit_size=1, graph coordinates map directly to visual positions.
-   * Applies Manim→THREE.js mapping: (mx, my, mz) → (mx, mz, -my)
    */
   coordsToPoint(x: number, y: number, z: number): Vector3Tuple {
```
maloyan, line 373: **Medium — silent public API break.** `coordsToPoint / pointToCoords / c2p{X,Y,Z}` and the tick/label positions all now return different values for the same inputs than `main` (old Manim→THREE mapping). There is no CHANGELOG entry, migration note, or PR-body callout. Any user scene that placed mobjects via `coordsToPoint()` or asserted `getXLabel().position` will break at runtime with no warning. Add an explicit migration note and bump semver-major, or ship behind a `convention: "manim" | "three"` switch with a deprecation warning on the old path.

examples/3d-scenes/orbit_controls_basis_experiment.html
```diff
@@ -0,0 +1,149 @@
...
+      log(`${label}
+cam.up=${fmt(up)}
+storedQuatVsRefAngle=${Number.isFinite(quatDelta) ? quatDelta.toFixed(6) : 'n/a'}
+map(up)->Y error=${Number.isFinite(mappedErr) ? mappedErr.toExponential(3) : 'n/a'}
+pos=${fmt(cam.position)} target=${fmt(controls.target)} lockX=${lockX}`);
+    }
+
+    function rebuildQuatFromCurrentUp() {
+      const up = cam.up.clone().normalize();
+      controls._quat = new THREE.Quaternion().setFromUnitVectors(up, Y);
```
maloyan: **Low — example pokes private three.js internals.** `controls._quat` / `controls._quatInverse` are not part of the `OrbitControls` public API and will break on any routine three.js update. Shipping this under `examples/` bakes in an unsupported contract. Either move to a non-shipping debug/scratch area (e.g. `tools/` or `dev/`), or expose a supported basis-rebuild method on our own `OrbitControls` wrapper and have the example use that.

examples/3d-scenes/three_d_axes_coords.html
```diff
@@ -0,0 +1,72 @@
...
+    const scene = new ThreeDScene(document.getElementById('container'), {
+      width: window.innerWidth,
+      height: window.innerHeight,
+      backgroundColor: '#191919',
+      phi: 0.0,
+      theta: -90 * (Math.PI / 180),
+      distance: 20,
+      fov: 30,
+      enableOrbitControls: true,
+      orbitControlsUp: "x",
```
maloyan: **Low — confusing config for a #256 regression demo.** `phi: 0` puts the camera on +Z, but `orbitControlsUp: "x"` makes screen-up +X, so the visual check of "coordsToPoint(1,0,0) sits on the X axis" is harder to reason about and exercises the special x-up path instead of the default basis. If this is meant to validate axis alignment, keep the default camera basis here and move x-up behavior into a separate demo.

rambip:
fixed

PR #285, opened 2026-04-23
fix: fill mesh geometry includes Z coordinates for 3D rotation

rambip:
## Summary

When rotating a VMobject (like Circle) in 3D, and using camera orbit conrols. the contour (stroke) would rotate correctly but the fill mesh would not. This was because:

1. sampleBezierOutline() dropped Z coordinates, returning only [x, y]
2. buildEarcutFillGeometry() hardcoded Z=0 in the output positions

## Changes

Fix by preserving Z coordinates throughout the fill geometry pipeline:
- sampleBezierOutline() now returns [x, y, z] points
- buildEarcutFillGeometry() uses z from sampled outline
- buildEarcutFillGeometryMulti() uses z from vertices

Add test verifying both stroke and fill mesh have correct Z values after 3D rotation.

## Testing
- [x] Tests pass (`npm test`)

## Related Issues
fixes this:

image

### DIFF
--- a/src/core/VMobjectGeometry.ts
+++ b/src/core/VMobjectGeometry.ts
@@ -119,14 +119,14 @@ export function sampleBezierOutline(points: number[][], samplesPerSegment: numbe
     for (let t = startT; t <= samples; t++) {
       const u = t / samples;
       const pt = evalCubicBezier(p0, p1, p2, p3, u);
-      result.push([pt[0], pt[1]]);
+      result.push([pt[0], pt[1], pt[2] ?? 0]); // Include Z for 3D support
     }
   }
 
   // Handle non-Bezier (simple line segment) fallback
   if (result.length === 0 && points.length >= 2) {
     for (const p of points) {
-      result.push([p[0], p[1]]);
+      result.push([p[0], p[1], p[2] ?? 0]); // Include Z for 3D support
     }
   }
 
@@ -314,7 +314,7 @@ export function buildEarcutFillGeometry(
     const v = outline[indices[i]];
     positions[i * 3] = v[0];
     positions[i * 3 + 1] = v[1];
-    positions[i * 3 + 2] = 0;
+    positions[i * 3 + 2] = v[2] ?? 0; // Use Z for 3D support
   }
 
   const geometry = new THREE.BufferGeometry();
@@ -387,7 +387,7 @@ function buildEarcutFillGeometryMulti(
 
     for (const idx of indices) {
       const v = allVerts[idx];
-      allPositions.push(v[0], v[1], 0);
+      allPositions.push(v[0], v[1], v[2] ?? 0); // Use Z for 3D support
     }
   }
 
--- a/src/core/VMobjectGeometry.ts
+++ b/src/core/VMobjectGeometry.ts
@@ -152,7 +152,7 @@ export function sampleBezierOutline(points: number[][], samplesPerSegment: numbe
  * Sample Bezier path into 3D points (preserves z).
  * Same logic as sampleBezierOutline but returns [x, y, z] instead of [x, y].
  */
-export function sampleBezierOutline3D(points: number[][], samplesPerSegment: number): number[][] {
+function sampleBezierOutline3D(points: number[][], samplesPerSegment: number): number[][] {
   const result: number[][] = [];
 
   for (let i = 0; i + 3 < points.length; i += 3) {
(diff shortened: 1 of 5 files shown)


maloyan:
## Codex review summary

**High**: preserving per-vertex Z only helps _after_ earcut succeeds. `triangulatePolygon()` and `pointInPolygon()` are still 2D (XY), so polygons whose XY projection is degenerate (exact XZ/YZ plane, near-quarter-turn rotations) will either fall back to `ShapeGeometry` at `z=0` or mis-classify holes in compound glyphs. Fix by projecting each ring onto a plane-local 2D basis for triangulation/containment, then emitting the original 3D vertices.
**Medium**: `buildEarcutFillGeometryMulti` hole ownership uses `pointInPolygon(rings[i][0], rings[j])` on raw XY — rotated compound glyphs can collapse these rings and fill counters solid.
**Low**: `sampleBezierOutline` docstring still says "returns [x, y] pairs (no z)" — stale.
**Low**: regression test only covers a shallow X-tilt on a single-ring circle; add an XZ/YZ-plane case and a multi-subpath shape (annulus / glyph counter) to catch the remaining projection bug.

**Positive**: propagating source `z` into both single- and multi-ring fill vertices is the right direction for non-degenerate projections; the new regression test is a useful start.

Inline comments on each site. _(Review via codex.)_

src/core/VMobjectGeometry.ts
```diff
@@ -314,7 +314,7 @@ export function buildEarcutFillGeometry(
     const v = outline[indices[i]];
     positions[i * 3] = v[0];
     positions[i * 3 + 1] = v[1];
-    positions[i * 3 + 2] = 0;
+    positions[i * 3 + 2] = v[2] ?? 0; // Use Z for 3D support
```
maloyan: **High — fix is incomplete for plane-degenerate projections.** `positions[i * 3 + 2] = v[2] ?? 0` only preserves depth _after_ earcut succeeds in XY. `triangulatePolygon()` still discards Z, so polygons whose XY projection is degenerate (exact XZ/YZ-plane ring, or a transform that snaps there) still hit the `ShapeGeometry` fallback above and get rebuilt in the XY plane at `z=0`. Triangulate in a plane-local 2D basis derived from the polygon normal, then map indices back to the original 3D vertices.

src/core/VMobjectGeometry.ts
```diff
@@ -387,7 +387,7 @@ function buildEarcutFillGeometryMulti(
 
     for (const idx of indices) {
       const v = allVerts[idx];
-      allPositions.push(v[0], v[1], 0);
+      allPositions.push(v[0], v[1], v[2] ?? 0); // Use Z for 3D support
```
maloyan: **Medium — hole detection stays 2D.** `buildEarcutFillGeometryMulti()` determines hole ownership with `pointInPolygon(rings[i][0], rings[j])` in XY only (above). For rotated compound glyphs whose projected rings collapse or become numerically skinny, that test can return false and the inner contour is treated as another outer ring, filling counters solid. Run containment on the same plane-projected coordinates you use for triangulation.

src/core/VMobjectGeometry.ts
```diff
@@ -119,14 +119,14 @@ export function sampleBezierOutline(points: number[][], samplesPerSegment: numbe
     for (let t = startT; t <= samples; t++) {
       const u = t / samples;
       const pt = evalCubicBezier(p0, p1, p2, p3, u);
-      result.push([pt[0], pt[1]]);
+      result.push([pt[0], pt[1], pt[2] ?? 0]); // Include Z for 3D support
```
maloyan, line 128: **Low — stale docstring.** `sampleBezierOutline()` now returns 3-tuples, but the docblock above (line ~112-114) still says it "returns [x, y] pairs (no z) and skips duplicate-point de-duplication...". Update the contract comment so downstream callers do not keep assuming a 2D-only return shape. Also: the closing-duplicate check near the end only compares X/Y — consider whether a Z-varying loop endpoint could be falsely deduped (probably fine for closed rings on a single plane, but worth a note).

src/core/circle-rotate-3d.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const circle = new Circle({
+      radius: 1,
+      color: '#ff0000',
+      fillOpacity: 0.7,
+      strokeWidth: 2,
+    });
+
+    // Force initial geometry build
+    circle.getThreeObject();
+
```
maloyan, line 18: **Low — test gaps.** This regression only exercises a shallow X-axis tilt (`Math.PI / 12`) on a single-ring Circle, so it will not catch (a) the remaining projection bug flagged on `buildEarcutFillGeometry` when the XY projection is degenerate, or (b) the compound-hole case in `buildEarcutFillGeometryMulti`. Add cases for: near-quarter-turn / exact XZ-plane rotation, rotation around Y and Z axes, and a multi-subpath shape (annulus, glyph counter) so hole classification is covered.

PR #298, opened 2026-04-27
fix(transform): preserve compound path metadata at finish

### LINKED ISSUE #280: When MathTex is used with a transform that changes the number of glyphs, end state is broken
## Description

https://github.com/user-attachments/assets/167ed63a-6078-4c5a-9c7b-5343d85f014e

## Steps to Reproduce
play a transform from $x=1$ to $x=42$

## Expected Behavior
What you expected to happen.

## Actual Behavior
What actually happened.

## Environment
- Browser:
- OS:
- manim-web version:

## Additional Context
The fix may be as simple as replacing the last state by the target MathTex state.

rambip:
# Summary

Ensures Transform ends with exact target geometry/topology for compound glyphs (including holes), and fixes rough/incorrect fill outlines for MathTex compound glyphs by preserving fill-critical curve sampling.
This resolves final-frame hole artifacts and the curved-ring undersampling issue that caused visibly faceted fills.

# Changes

- Restore exact target subpath lengths at finish() for single VMobject transforms.
- Restore exact target subpath lengths per child for VGroup transforms.
- Use typed transform-time subpath override (setTransformSubpathLengths) during interpolation, then clear it at finish.
- Remove interpolate-time metadata sync fallback.
- Add strict validation in compound alignment (sum(lengths) === points.length).
- Improve alignment quality path with arc-length resampling and minimum 64 points/subpath.
- Fix fill sampling for triangulation by removing near-linear collapse in fill outline sampling (uniform per-segment sampling for fill path).
- Introduce a dedicated fill sampling constant (FILL_SAMPLES_PER_SEGMENT) instead of hardcoded values.
- Refactor subpath metadata ownership: replace SVG monkey-patched getSubpathLengths with explicit VMobject base metadata (setBaseSubpathLengths / getSubpathLengths) and keep getEffectiveSubpathLengths as resolver.
- Remove temporary debug instrumentation and clean up unused exports in geometry helpers.
- Revert the attempted cross-frame earcut index reuse optimization due to end-state triangulation correctness regressions.

# Testing

- Targeted tests pass:
    - src/core/VMobjectGeometry.test.ts
    - src/animation/transform.test.ts
    - src/animation/transform/TransformPairing.test.ts
    - src/animation/animation-transform-extra.test.ts
- Typecheck passes.
- Added/updated assertions confirm:
    - transform-time aligned lengths (e.g. [64, 64])
    - finish-time exact target lengths restored (e.g. [5, 5])
    - dense curved-ring sampling retained for fill triangulation regression case.

# Related Issues

Closes #280.

# Optimization

The performance optimization that reused earcut indices across frames was reverted because it could produce incorrect final triangulation when topology changed subtly.
If we revisit performance, we should prioritize safe wins (e.g., skipping rebuild only when geometry metadata is unchanged) rather than cross-frame index reuse heuristics.


rambip:
#  Goal                                                                                                                                                                                      
                                                                                                                                                                                           
 Evaluate mid-transform triangulation fixes in two flavors, starting from a clean baseline where initial frame of source glyph is correct (your key observation).                          
                                                                                                                                                                                           
## Precondition / sanity check first                                                                                                                                                         
                                                                                                                                                                                           
 Before any transform changes, verify:                                                                                                                                                     
 1. MathTex("2") rendered alone has correct fill/holes at rest.                                                                                                                            
 2. Same MathTex("2") immediately before scene.play(Transform(...)) is still correct.                                                                                                      
 3. If broken already at start in repro, isolate whether repro itself mutates source (debug code or copied metadata too early).                                                            
                                                                                                                                                                                           
 If precondition fails, fix repro/instrumentation first; otherwise transform tests are invalid.                                                                                            
                                                                                                                                                                                           
 ────────────────────────────────────────────────────────────────────────────────                                                                                                          
                                                                                                                                                                                           
 ## Flavor A — “midway better than nothing” (minimal experiment)                                                                                                                              
                                                                                                                                                                                           
 Purpose: test if re-triangulating from evolving points during transform reduces artifacts versus only finish-time correction.                                                             
                                                                                                                                                                                           
 ### A1. Trigger strategy                                                                                                                                                                  
                                                                                                                                                                                           
 - In Transform.interpolate path, for transformed VMobject(s), periodically force fill geometry rebuild (e.g. every N frames or alpha delta threshold), not just at end.                   
 - No async/worker; stay sync main-thread.                                                                                                                                                 
                                                                                                                                                                                           
 ### A2. Scope                                                                                                                                                                             
                                                                                                                                                                                           
 - Start with glyphs that expose compound semantics (getSubpaths available).                                                                                                               
 - Apply to:                                                                                                                                                                               
     - single VMobject transform                                                                                                                                                           
     - VGroup matched child transforms                                                                                                                                                     
 - Do not add broader topology heuristics yet.                                                                                                                                             
                                                                                                                                                                                           
 ### A3. Evaluation                                                                                                                                                                        
                                                                                                                                                                                           
 - Visual compare in repro:                                                                                                                                                                
     1. baseline (current finish-only fix)                                                                                                                                                 
     2. flavor A enabled                                                                                                                                                                   
 - Success criterion: fewer obvious hole/fill flips during mid-frames, while initial source frame remains correct.                                                                         
                                                                                                                                                                                           
 ────────────────────────────────────────────────────────────────────────────────                                                                                                          
                                                                                                                                                                                           
## Flavor B — “triangle invasion” mitigation (topology stability)                                                                                                                            
                                                                                                                                                                                           
 Purpose: specifically address invasive/flickering triangulation when holes/contours move across each other during interpolation.                                                          
                                                                                                                                                                                           
 ### B1. Add contour-aware guardrails                                                                                                                                                      
                                                                                                                                                                                           
 - At retriangulation points, use contour classification from getSubpaths:                                                                                                                 
     - enforce stable outer/inner ordering                                                                                                                                                 
     - reject/recover from invalid hole assignment for current alpha                                                                                                                       
 - If classification becomes ambiguous at a frame, choose deterministic fallback (e.g. previous valid triangulation) to avoid sudden invasions.                                            
                                                                                                                                                                                           
 ### B2. Optional cadence                                                                                                                                                                  
                                                                                                                                                                                           
 - Run B only when cheap detector says topology risk increased (e.g. winding/containment flips), otherwise use A cadence.                                                                  
                                                                                                                                                                                           
 ### B3. Evaluation                                                                                                                                                                        
                                                                                                                                                                                           
 - Same repro, slow duration.                                                                                                                                                              
 - Track:                                                                                                                                                                                  
     - start frame correctness (must remain correct)                                                                                                                                       
     - count of visible fill inversions/invasions through transition                                                                                                                       
     - frame-time spikes (keep acceptable)                                                                                                                                                 
                                                                                                                                                                                           
 ────────────────────────────────────────────────────────────────────────────────                                                                                                          
                                                                                                                                                                                           
 ## Rollout order                                                                                                                                                                             
                                                                                                                                                                                           
 1. Fix/lock repro so start state of "2" is definitely correct.                                                                                                                            
 2. Implement Flavor A minimal hook; validate “better than nothing.”                                                                                                                       
 3. Only if A helps but still invades, implement Flavor B topology stabilization.                                                                                                          
 4. Keep all work out of examples commit unless explicitly requested.

rambip:
**making progress**:

Before:


https://github.com/user-attachments/assets/ab9e33c1-b457-4568-90b0-96655d745d5d

Now:


https://github.com/user-attachments/assets/c8e21d58-c4c7-4afd-bc61-5715b5a367b8

rambip:
@maloyan could you start to review please ?

rambip:
# Implementation Plan                                                                                                                                                                       
                                                                                                                                                                                           
 ### 1. Core Function: extractLeafPairsWithWorld                                                                                                                                           
                                                                                                                                                                                           
 Location: src/core/MobjectTraversal.ts (rename or extend current file)                                                                                                                    
                                                                                                                                                                                           
 Signature:                                                                                                                                                                                
                                                                                                                                                                                           
 ```typescript                                                                                                                                                                             
   interface LeafPair {                                                                                                                                                                    
     src: VMobject | null;                                                                                                                                                                 
     tgt: VMobject | null;                                                                                                                                                                 
     srcWorldMatrix: THREE.Matrix4 | null;                                                                                                                                                 
     tgtWorldMatrix: THREE.Matrix4 | null;                                                                                                                                                 
   }                                                                                                                                                                                       
                                                                                                                                                                                           
   function extractLeafPairsWithWorld(                                                                                                                                                     
     src: VMobject | null,                                                                                                                                                                 
     tgt: VMobject | null,                                                                                                                                                                 
     srcParentWorldMatrix: THREE.Matrix4,                                                                                                                                                  
     tgtParentWorldMatrix: THREE.Matrix4                                                                                                                                                   
   ): LeafPair[]                                                                                                                                                                           
 ```                                                                                                                                                                                       
                                                                                                                                                                                           
 Implementation:                                                                                                                                                                           
                                                                                                                                                                                           
 ```typescript                                                                                                                                                                             
   function extractLeafPairsWithWorld(                                                                                                                                                     
     src: VMobject | null,                                                                                                                                                                 
     tgt: VMobject | null,                                                                                                                                                                 
     srcParentMatrix: THREE.Matrix4,                                                                                                                                                       
     tgtParentMatrix: THREE.Matrix4                                                                                                                                                        
   ): LeafPair[] {                                                                                                                                                                         
     // Compute world matrices for current nodes                                                                                                                                           
     const srcWorldMatrix = src ? composeWorldMatrix(src, srcParentMatrix) : null;                                                                                                         
     const tgtWorldMatrix = tgt ? composeWorldMatrix(tgt, tgtParentMatrix) : null;                                                                                                         
                                                                                                                                                                                           
     // Both null - shouldn't happen                                                                                                                                                       
     if (!src && !tgt) return [];                                                                                                                                                          
                                                                                                                                                                                           
     // Fade in: src is null, extract all tgt leaves                                                                                                                                       
     if (!src) {                                                                                                                                                                           
       return getLeavesWithMatrix(tgt!, tgtWorldMatrix!).map(leaf => ({                                                                                                                    
         src: null,                                                                                                                                                                        
         tgt: leaf.vmobject,                                                                                                                                                               
         srcWorldMatrix: null,                                                                                                                                                             
         tgtWorldMatrix: leaf.worldMatrix,                                                                                                                                                 
       }));                                                                                                                                                                                
     }                                                                                                                                                                                     
                                                                                                                                                                                           
     // Fade out: tgt is null, extract all src leaves                                                                                                                                      
     if (!tgt) {                                                                                                                                                                           
       return getLeavesWithMatrix(src, srcWorldMatrix!).map(leaf => ({                                                                                                                     
         src: leaf.vmobject,                                                                                                                                                               
         tgt: null,                                                                                                                                                                        
         srcWorldMatrix: leaf.worldMatrix,                                                                                                                                                 
         tgtWorldMatrix: null,                                                                                                                                                             
       }));                                                                                                                                                                                
     }                                                                                                                                                                                     
                                                                                                                                                                                           
     // Both exist - check if leaves                                                                                                                                                       
     const srcChildren = getVMobjectChildren(src);                                                                                                                                         
     const tgtChildren = getVMobjectChildren(tgt);                                                                                                                                         
                                                                                                                                                                                           
     const srcIsLeaf = srcChildren.length === 0;                                                                                                                                           
     const tgtIsLeaf = tgtChildren.length === 0;                                                                                                                                           
                                                                                                                                                                                           
     // Both leaves - base case                                                                                                                                                            
     if (srcIsLeaf && tgtIsLeaf) {                                                                                                                                                         
       return [{                                                                                                                                                                           
         src,                                                                                                                                                                              
         tgt,                                                                                                                                                                              
         srcWorldMatrix: srcWorldMatrix!,                                                                                                                                                  
         tgtWorldMatrix: tgtWorldMatrix!,                                                                                                                                                  
       }];                                                                                                                                                                                 
     }                                                                                                                                                                                     
                                                                                                                                                                                           
     // At least one is a group - align children and recurse                                                                                                                               
     const srcItems = srcIsLeaf ? [src] : srcChildren;                                                                                                                                     
     const tgtItems = tgtIsLeaf ? [tgt] : tgtChildren;                                                                                                                                     
                                                                                                                                                                                           
     const maxLen = Math.max(srcItems.length, tgtItems.length);                                                                                                                            
     const pairs: LeafPair[] = [];                                                                                                                                                         
                                                                                                                                                                                           
     for (let i = 0; i < maxLen; i++) {                                                                                                                                                    
       pairs.push(...extractLeafPairsWithWorld(                                                                                                                                            
         srcItems[i] ?? null,                                                                                                                                                              
         tgtItems[i] ?? null,                                                                                                                                                              
         srcWorldMatrix!,                                                                                                                                                                  
         tgtWorldMatrix!                                                                                                                                                                   
       ));                                                                                                                                                                                 
     }                                                                                                                                                                                     
                                                                                                                                                                                           
     return pairs;                                                                                                                                                                         
   }                                                                                                                                                                                       
                                                                                                                                                                                           
   // Helper: get direct VMobject children, or empty array for leaves                                                                                                                      
   function getVMobjectChildren(mob: VMobject): VMobject[] {                                                                                                                               
     if (mob instanceof VGroup) {                                                                                                                                                          
       return mob.children.filter(c => c instanceof VMobject) as VMobject[];                                                                                                               
     }                                                                                                                                                                                     
     return [];                                                                                                                                                                            
   }                                                                                                                                                                                       
                                                                                                                                                                                           
   // Helper: get all leaves with their world matrices                                                                                                                                     
   function getLeavesWithMatrix(                                                                                                                                                           
     mob: VMobject,                                                                                                                                                                        
     parentWorldMatrix: THREE.Matrix4                                                                                                                                                      
   ): { vmobject: VMobject; worldMatrix: THREE.Matrix4 }[] {                                                                                                                               
     const worldMatrix = composeWorldMatrix(mob, parentWorldMatrix);                                                                                                                       
                                                                                                                                                                                           
     const children = getVMobjectChildren(mob);                                                                                                                                            
     if (children.length === 0) {                                                                                                                                                          
       return [{ vmobject: mob, worldMatrix }];                                                                                                                                            
     }                                                                                                                                                                                     
                                                                                                                                                                                           
     return children.flatMap(child => getLeavesWithMatrix(child, worldMatrix));                                                                                                            
   }                                                                                                                                                                                       
                                                                                                                                                                                           
   // Helper: compose local transform with parent world matrix                                                                                                                             
   function composeWorldMatrix(mob: VMobject, parentWorldMatrix: THREE.Matrix4): THREE.Matrix4 {                                                                                           
     const local = new THREE.Matrix4();                                                                                                                                                    
     const q = new THREE.Quaternion().setFromEuler(mob.rotation);                                                                                                                          
     local.compose(mob.position, q, mob.scaleVector);                                                                                                                                      
     return new THREE.Matrix4().multiplyMatrices(parentWorldMatrix, local);                                                                                                                
   }                                                                                                                                                                                       
 ```                                                                                                                                                                                       
                                                                                                                                                                                           
 ────────────────────────────────────────────────────────────────────────────────                                                                                                          
                                                                                                                                                                                           
 ### 2. Transform.begin(): Setup Leaf Pairs                                                                                                                                                
                                                                                                                                                                                           
 ```typescript                                                                                                                                                                             
   override begin(): void {                                                                                                                                                                
     super.begin();                                                                                                                                                                        
                                                                                                                                                                                           
     const identity = new THREE.Matrix4().identity();                                                                                                                                      
     this._leafPairs = extractLeafPairsWithWorld(                                                                                                                                          
       this.mobject as VMobject,                                                                                                                                                           
       this.target as VMobject,                                                                                                                                                            
       identity,                                                                                                                                                                           
       identity                                                                                                                                                                            
     );                                                                                                                                                                                    
                                                                                                                                                                                           
     // Setup interpolation state for each pair                                                                                                                                            
     this._leafStates = this._leafPairs.map(pair => this._setupLeafState(pair));                                                                                                           
   }                                                                                                                                                                                       
                                                                                                                                                                                           
   interface LeafState {                                                                                                                                                                   
     // Points                                                                                                                                                                             
     startPoints: number[][];                                                                                                                                                              
     targetPoints: number[][];                                                                                                                                                             
     finalTargetPoints: number[][];  // Unaligned, exact target                                                                                                                            
     finalSubpathLengths: number[] | undefined;                                                                                                                                            
                                                                                                                                                                                           
     // Position (world space projected to source parent)                                                                                                                                  
     startPosition: THREE.Vector3;                                                                                                                                                         
     targetPosition: THREE.Vector3;                                                                                                                                                        
                                                                                                                                                                                           
     // Style                                                                                                                                                                              
     startOpacity: number;                                                                                                                                                                 
     targetOpacity: number;                                                                                                                                                                
     // ... other style props                                                                                                                                                              
   }                                                                                                                                                                                       
                                                                                                                                                                                           
   private _setupLeafState(pair: LeafPair): LeafState {                                                                                                                                    
     if (pair.src && pair.tgt) {                                                                                                                                                           
       // Normal transform                                                                                                                                                                 
       const srcCopy = pair.src.copy() as VMobject;                                                                                                                                        
       const tgtCopy = pair.tgt.copy() as VMobject;                                                                                                                                        
                                                                                                                                                                                           
       const aligned = alignCompoundPathsForTransform(                                                                                                                                     
         srcCopy.getPoints(),                                                                                                                                                              
         srcCopy.getEffectiveSubpathLengths?.(),                                                                                                                                           
         tgtCopy.getPoints(),                                                                                                                                                              
         tgtCopy.getEffectiveSubpathLengths?.()                                                                                                                                            
       );                                                                                                                                                                                  
                                                                                                                                                                                           
       return {                                                                                                                                                                            
         startPoints: aligned?.srcAlignedPoints ?? srcCopy.getPoints(),                                                                                                                    
         targetPoints: aligned?.tgtAlignedPoints ?? tgtCopy.getPoints(),                                                                                                                   
         finalTargetPoints: pair.tgt.getPoints(),                                                                                                                                          
         finalSubpathLengths: pair.tgt.getEffectiveSubpathLengths?.(),                                                                                                                     
         startPosition: extractPosition(pair.srcWorldMatrix),                                                                                                                              
         targetPosition: extractPosition(pair.tgtWorldMatrix),                                                                                                                             
         startOpacity: pair.src.opacity,                                                                                                                                                   
         targetOpacity: pair.tgt.opacity,                                                                                                                                                  
       };                                                                                                                                                                                  
     }                                                                                                                                                                                     
                                                                                                                                                                                           
     if (pair.src) {                                                                                                                                                                       
       // Fade out                                                                                                                                                                         
       return {                                                                                                                                                                            
         startPoints: pair.src.getPoints(),                                                                                                                                                
         targetPoints: pair.src.getPoints(),  // Same points                                                                                                                               
         finalTargetPoints: pair.src.getPoints(),                                                                                                                                          
         finalSubpathLengths: undefined,                                                                                                                                                   
         startPosition: extractPosition(pair.srcWorldMatrix!),                                                                                                                             
         targetPosition: extractPosition(pair.srcWorldMatrix!),  // Stay in place                                                                                                          
         startOpacity: pair.src.opacity,                                                                                                                                                   
         targetOpacity: 0,                                                                                                                                                                 
       };                                                                                                                                                                                  
     }                                                                                                                                                                                     
                                                                                                                                                                                           
     // Fade in (pair.tgt exists)                                                                                                                                                          
     const placeholder = createPlaceholderFor(pair.tgt!);                                                                                                                                  
     (this.mobject as VGroup).add(placeholder);                                                                                                                                            
     pair.src = placeholder;  // Mutate pair to reference placeholder                                                                                                                      
                                                                                                                                                                                           
     return {                                                                                                                                                                              
       startPoints: placeholder.getPoints(),                                                                                                                                               
       targetPoints: pair.tgt.getPoints(),                                                                                                                                                 
       finalTargetPoints: pair.tgt.getPoints(),                                                                                                                                            
       finalSubpathLengths: pair.tgt.getEffectiveSubpathLengths?.(),                                                                                                                       
       startPosition: extractPosition(pair.tgtWorldMatrix!),                                                                                                                               
       targetPosition: extractPosition(pair.tgtWorldMatrix!),                                                                                                                              
       startOpacity: 0,                                                                                                                                                                    
       targetOpacity: pair.tgt.opacity,                                                                                                                                                    
     };                                                                                                                                                                                    
   }                                                                                                                                                                                       
 ```                                                                                                                                                                                       
                                                                                                                                                                                           
 ────────────────────────────────────────────────────────────────────────────────                                                                                                          
                                                                                                                                                                                           
 ### 3. Interpolation Logic                                                                                                                                                                
                                                                                                                                                                                           
 ```typescript                                                                                                                                                                             
     interpolate(alpha: number): void {                                                                                                                                                      
     for (let i = 0; i < this._leafPairs.length; i++) {                                                                                                                                    
       const pair = this._leafPairs[i];                                                                                                                                                    
       const state = this._leafStates[i];                                                                                                                                                  
                                                                                                                                                                                           
       if (!pair.src) continue;                                                                                                                                                            
                                                                                                                                                                                           
       // Interpolate points                                                                                                                                                               
       pair.src.setPoints(lerpPoints(state.startPoints, state.targetPoints, alpha));                                                                                                       
                                                                                                                                                                                           
       // Interpolate position - just set the interpolated world position                                                                                                                  
       pair.src.position.lerpVectors(state.startPosition, state.targetPosition, alpha);                                                                                                    
                                                                                                                                                                                           
       // Interpolate style                                                                                                                                                                
       pair.src.opacity = lerp(state.startOpacity, state.targetOpacity, alpha);                                                                                                            
                                                                                                                                                                                           
       pair.src._markDirty();                                                                                                                                                              
     }                                                                                                                                                                                     
   }                                                                                                                                                                           
 ```                                                                                                                                                                                       
                                                                                                                                                                                           
 ────────────────────────────────────────────────────────────────────────────────                                                                                                          
                                                                                                                                                                                           
 ### 4. Finish(): Restore Target State                                                                                                                                                     
                                                                                                                                                                                           
 ```typescript                                                                                                                                                                             
    finish(): void {     
    // Note we have to make sure we match the target state exactly. Maybe it will be better to copy the original target state and use it there if it's too complicated to match exactly.                                                                                                                     
     for (const pair of this._leafPairs) {                                                                                                                                                 
       if (pair.src && pair.tgt) {                                                                                                                                                         
         // Copy exact state from original target reference                                                                                                                                
         pair.src.setPoints(pair.tgt.getPoints());                                                                                                                                         
         pair.src.setTransformSubpathLengths(pair.tgt.getEffectiveSubpathLengths?.());                                                                                                     
         pair.src.opacity = pair.tgt.opacity;                                                                                                                                              
         pair.src.color = pair.tgt.color;                                                                                                                                                  
         // ... other style props                                                                                                                                                          
         pair.src._markDirty();                                                                                                                                                            
       }                                                                                                                                                                                   
     }                                                                                                                                                                                     
     super.finish();                                                                                                                                                                       
   }                                                                                                                                        
 ```                                                                                                                                                                                       
                                                                                                                                                                                           
 ────────────────────────────────────────────────────────────────────────────────                                                                                                          
                                                                                                                                                                                           
 ### 5. Cleanup / Edge Cases                                                                                                                                                               
                                                                                                                                                                                           
 - Placeholder management: Track fade-in placeholders to remove them after animation                                                                                                       
 - Empty VGroups: getVMobjectChildren returns [], treated same as leaf for alignment                                                                                                       
 - Deep nesting: Recursion handles arbitrary depth                                                                                                                                         
 - World position accuracy: Matrices computed during traversal, passed down recursively                                                                                                    
                                                                                                                                                                                           
 ────────────────────────────────────────────────────────────────────────────────                                                                                                          
                                                                                                                                                                                           
 ### Migration Path                                                                                                                                                                        
                                                                                                                                                                                           
 1. Add extractLeafPairsWithWorld to MobjectTraversal.ts                                                                                                                                   
 2. Add _leafPairs and _leafStates to Transform.ts                                                                                                                                         
 3. Replace _beginVGroup logic with new approach                                                                                                                                           
 4. Update interpolate and finish to use leaf pairs                                                                                                                                        
 5. Remove old _childXxx arrays

src/animation/transform/Transform.ts
```diff
@@ -46,6 +46,23 @@ interface ChildStyle {
   fillColor: string;
 }
 
+interface CompoundPathVMobject extends VMobject {
+  getSubpaths?: () => number[];
```
maloyan: **Critical — name collision with `SVGMobject.getSubpaths()`.**

`SVGMobject` (`src/mobjects/svg/SVGMobject.ts:841`) defines `getSubpaths(): VMobject[]` on the prototype — same method name, completely different return type. When `target` is an `SVGMobject`, `targetCompound.getSubpaths` is truthy → bound → receiver gets `() => [...VMobject[]]` typed as `number[]`. Renderer at `src/core/VMobjectRendering.ts:205,320` and `src/core/VMobjectGeometry.ts:403` then feeds VMobject objects into earcut as if they were lengths → silent NaN/garbage triangulation, possibly worse than #280.

Cannot fix structurally with this interface alone. Either rename the numeric API to `getSubpathLengths()`, or move metadata to a `WeakMap<VMobject, number[]>` keyed off the receiver (Codex's suggested fix in the review body).

src/animation/transform/Transform.ts
```diff
@@ -46,6 +46,23 @@ interface ChildStyle {
...
+}
+
+/** Keep compound-path metadata (holes/subpaths) in sync with target geometry. */
+function syncCompoundPathMetadata(target: VMobject, receiver: VMobject): void {
+  const targetCompound = target as CompoundPathVMobject;
+  const receiverCompound = receiver as CompoundPathVMobject;
+
+  if (targetCompound.getSubpaths) {
+    const getSubpaths = targetCompound.getSubpaths.bind(targetCompound);
+    receiverCompound.getSubpaths = () => [...getSubpaths()];
```
maloyan: **Closure binds the live target.** `bind(targetCompound)` keeps `target` GC-pinned for the receiver's lifetime, and any later mutation of target's internal subpath array silently propagates to the receiver — minutes after the Transform finished.

Snapshot the values now instead of binding:
```ts
const snap = [...targetCompound.getSubpaths()];
receiverCompound.getSubpaths = () => [...snap];
```
This is also more consistent with the rest of `finish()`, which copies static final state, not live references.

src/animation/transform/Transform.ts
```diff
@@ -46,6 +46,23 @@ interface ChildStyle {
...
+  } else {
+    delete receiverCompound.getSubpaths;
```
maloyan: **Critical — `delete` is a no-op for prototype methods.** Three classes own `getSubpaths` on their prototype:
- `Cutout` — `src/mobjects/geometry/PolygonExtensions.ts:844`
- `BooleanResult` — `src/mobjects/geometry/BooleanOperations.ts:67`
- `SVGMobject` — `src/mobjects/svg/SVGMobject.ts:841`

Scenario: source is a `Cutout` (prototype `getSubpaths`), target is a plain `Polygon` (no `getSubpaths`). This branch runs `delete` on a property that doesn't exist on the instance. The prototype method survives → receiver still reports stale subpath lengths from the deleted hole → renderer slices flat polygon points using stale subpath counts → silently corrupted geometry.

Fix options:
1. Overwrite as own property: `receiverCompound.getSubpaths = () => []` (or a sentinel that the renderer knows to ignore).
2. Move metadata off the public method entirely (WeakMap approach in the review body).

src/animation/transform/Transform.ts
```diff
@@ -499,16 +537,25 @@ export class Transform extends Animation {
...
         child.position.copy(this._childTargetPositions[c]);
         const ts = this._childTargetStyles[c];
         child.opacity = ts.opacity;
         child.fillOpacity = ts.fillOpacity;
         child.strokeWidth = ts.strokeWidth;
         child.color = ts.color;
         child.fillColor = ts.fillColor;
+
+        const metadataSource = targetChild ?? child;
+        syncCompoundPathMetadata(metadataSource, child);
```
maloyan: **Orphan-source self-sync layers closures across re-transforms.** When `targetChild` is `undefined` (the fade-out branch at line 305), `metadataSource = child` → this calls `syncCompoundPathMetadata(child, child)`. Walking the helper with target === receiver: it binds the receiver's *current* `getSubpaths` (which may itself be a closure from a prior Transform), then re-assigns `receiver.getSubpaths = () => [...prevClosure()]`. Each Transform on the same orphan child layers another wrapper. Slow leak + stack-pointer hostile when the inner closure throws.

Guard the call:
```ts
if (targetChild) {
  syncCompoundPathMetadata(targetChild, child);
}
```
No-target child has nothing to sync from.

maloyan (commented):
Multi-agent + Codex review summary. Fix shape is right (snapshot pre-alignment target points + restore at finish), but the metadata-sync helper has two latent bugs and the PR body doesn't match what shipped.

**Critical**
- `CompoundPathVMobject.getSubpaths?: () => number[]` collides with `SVGMobject.getSubpaths(): VMobject[]` (`src/mobjects/svg/SVGMobject.ts:841`) — same name, incompatible return type. Renderer at `src/core/VMobjectRendering.ts:205,320` will receive `VMobject[]` typed as `number[]` → silent NaN/garbage triangulation when target is an SVG node.
- `delete receiverCompound.getSubpaths` is a no-op for prototype methods. `Cutout` (`PolygonExtensions.ts:844`), `BooleanResult` (`BooleanOperations.ts:67`), and `SVGMobject` all define `getSubpaths` on the prototype. Transforming any of them to a non-compound target leaves stale subpath data live.

**Important**
- Closure binds the live target (`bind(targetCompound)`) — keeps target alive and exposes the receiver to later target mutations. Snapshot the array at finish() instead.
- Orphan-source self-sync `syncCompoundPathMetadata(child, child)` (line 557 via `targetChild ?? child`) layers wrapper closures on each re-transform. Wrap in `if (targetChild)`.
- `_finalTargetPoints.length > 0 ? ... : _targetPoints` fallback at line 572 is unreachable today, but if a future edit hits it, the fix is silently bypassed (aligned points used instead of original). Use `null` sentinel + assert.

**PR description mismatch** — body mentions `setTransformSubpathLengths` typed override and removal of an interpolate-time fallback. Neither exists in the diff (`grep` for the symbol = 0 hits). Update the body, or implement the WeakMap-based version below — it's actually the cleaner fix.

**Zero tests** — fix targets a real visual bug (#280) but no assertion that `src.getSubpaths()` matches target post-finish, no `sum(subpathLengths) === points.length` topology invariant, no SVGMobject-target case. Body even admits "no visible effect" — likely because the demo doesn't exercise the compound branch.

**Suggested refactor** (Codex's recommendation — kills both critical bugs at once):
```ts
const transformSubpaths = new WeakMap<VMobject, number[] | null>();

function readSubpathLengths(m: VMobject): number[] | undefined {
  const raw = (m as unknown as { getSubpaths?: () => unknown }).getSubpaths?.();
  if (!Array.isArray(raw) || !raw.every(n => Number.isInteger(n) && n > 0)) return undefined;
  return raw.reduce((s, n) => s + n, 0) === m.getPoints().length ? [...raw] : undefined;
}

function syncCompoundPathMetadata(target: VMobject, receiver: VMobject): void {
  transformSubpaths.set(receiver, readSubpathLengths(target) ?? null);
}
```
Then renderer reads from `transformSubpaths` first, falling back to `getSubpaths()` for non-transformed mobjects. No prototype shadowing, no name collision, validated numeric shape, defensive snapshot.

**Nice to have** — nine parallel arrays indexed by child (now +2 from this PR) → `ChildSnapshot[]` struct would let TS enforce the lockstep invariant. Drop noise comments at lines 285, 305, 309, 327; keep WHY comments at 53, 198, 580.

src/animation/transform/Transform.ts
```diff
@@ -521,12 +568,17 @@ export class Transform extends Animation {
 
     const vmobject = this.mobject as VMobject;
     const vtarget = this.target as VMobject;
-    vmobject.setPoints(this._targetPoints);
+    vmobject.setPoints(
+      this._finalTargetPoints.length > 0 ? this._finalTargetPoints : this._targetPoints,
```
maloyan, line 541: **Unreachable fallback that silently defeats the fix.** The ternary `_finalTargetPoints.length > 0 ? final : target` is dead code today — every code path that reaches line 571 has already populated `_finalTargetPoints` at line 199. But if a future edit adds a non-VMobject target path or refactors `_beginVMobject`, this falls back to `_targetPoints` (the *aligned* version) without warning, and the entire purpose of this PR is silently undone.

Prefer a `null` sentinel + assert:
```ts
private _finalTargetPoints: number[][] | null = null;
// ...
if (this._finalTargetPoints === null) throw new Error('finish() before begin() captured _finalTargetPoints');
vmobject.setPoints(this._finalTargetPoints);
```
Length-zero-as-sentinel is the same anti-pattern as `null`-string-as-`""`.

src/animation/transform/Transform.ts
```diff
@@ -169,6 +195,8 @@ export class Transform extends Animation {
       const startCopy = vmobject.copy() as VMobject;
       const targetCopy = vtarget.copy() as VMobject;
 
+      // Keep final triangulation identical to the target by restoring this at finish().
+      this._finalTargetPoints = targetCopy.getPoints();
```
maloyan: Snapshot ordering is correct here — `targetCopy.getPoints()` is captured before `alignPoints()` mutates it. `getPoints()` returns a defensive copy (`VMobject.ts:101-103`) so the snapshot is safe.

Minor: the parallel field `_childOriginalTargetPoints` at line 280 captures `tc.getPoints()` (the *original* target child, not its copy). That works because `alignPoints` is called on `tcCopy` not `tc`, but the comment "BEFORE alignment" misleads — `tc` is never going to be mutated by alignment in either ordering. Reword to: `// Un-aligned target points; finish() restores these so triangulation matches the target exactly.`

rambip:
image

This will be a big PR. I'm trying to fix an old bug with triangulation of glyphs.

rambip:
Now: 

image

rambip:
https://github.com/user-attachments/assets/3abb934d-b821-498f-87e1-b639ee063072

I think that's good enough for now

Issue #260, opened 2026-04-16
Sliding object when trying to go beyond constraint


Wizorx:
## Description
Objects slide when trying to go beyond constraints set by `constrainX` or `constrainY` in `makeDraggable()`

## Steps to Reproduce
1. Create a draggable `Dot3D` (or some other object) with constraints in a given direction
2. Try dragging it past the constraint and it will slide in some other direction

## Expected Behavior
The object should not move if the user tries to move it beyond its constraint

## Actual Behavior
The object slides in some other direction

## Environment
- Browser: Brave 1.89.132
- OS: Windows 11 10.0.26200
- manim-web version: 0.3.19

## Additional Context
```<!DOCTYPE html>
<html>
<head>
  <style>
    * { margin: 0; padding: 0; box-sizing: border-box; }
    body { background: #191919; display: flex; justify-content: center; align-items: center; width: 100vw; height: 100vh; overflow: hidden; }
    #container { width: 100vw; height: 100vh; }
  </style>
</head>
<body>
  <div id="container"></div>

  <script type="module">
    import {ThreeDAxes, Dot3D, ThreeDScene, makeDraggable} from 'https://esm.sh/manim-web';

    const scene = new ThreeDScene(document.getElementById('container'), {
      width: window.innerWidth,
      height: window.innerHeight,
      backgroundColor: '#191919',
      phi: 75 * (Math.PI / 180),
      theta: -45 * (Math.PI / 180),
      distance: 20,
      fov: 30,
      enableOrbitControls: true,
    });

    const axes = new ThreeDAxes({
      xRange: [-5, 5, 1],
      yRange: [-5, 5, 1],
      zRange: [-5, 5, 1],
      xLength: 10,
      yLength: 10,
      zLength: 10,
      showLabels: true,
    });

    const dot = new Dot3D({ radius: 0.1}).moveTo(axes.coordsToPoint(1,1,1)).addUpdater(() => {});
    makeDraggable(dot, scene, {constrainX: [-5,5], constrainY: [-5,5]});


    scene.add(axes, dot);

    await scene.wait(Infinity);
  </script>
</body>
</html>```

maloyan:
## Investigation

Reproduced bug. Root cause in `src/interaction/Draggable.ts` `_updateDrag()`:

[...]

PR #304, opened 2026-05-01
feat(draggable): add constrainZ option (closes #261)

### LINKED ISSUE #261: Add constrainZ to makeDraggable()
## Description
An option to constrain draggable objects between certain values of the Z axis in a `ThreeDScene`.

## Use Case
It would help set up a "playable" zone in 3D scenes, and help with #260.

## Proposed Solution
Just like `constrainX` and `constrainY`.

maloyan:
## Summary
- Add `constrainZ?: [number, number] | null` to `DraggableOptions`, mirroring the existing `constrainX` / `constrainY` clamps.
- Apply the clamp after the existing #260 freeze-Z-on-X/Y-clamp logic so both behaviors compose cleanly.
- Extend `snapToGrid` to also snap Z in `ThreeDScene` (X/Y-only behavior is preserved in 2D scenes).
- Demo all three axes in `examples/3d-scenes/draggable_constraints_3d.{ts,html}`.

Closes #261.

## Test plan
- [x] `npx vitest run src/interaction/interaction.test.ts` — 94/94 pass, including new cases:
  - `constrainZ clamps Z within range when dragged past it`
  - `constrainZ works alongside constrainX and constrainY`
  - `constrainZ has no effect in 2D scenes (Z stays at mobject.position.z)` *(actually clamps the existing position.z value, see test)*
- [x] `npx vitest run` — full suite 5847/5847 pass.
- [x] `npx tsc --noEmit` — clean.
- [x] `npx prettier --check` + `npx eslint` on changed files — clean.
- [x] `npm run build` — clean.
- [x] Manual browser verification on `examples/3d-scenes/draggable_constraints_3d.html` with `{ constrainX: [-3,3], constrainY: [-3,3], constrainZ: [-3,3] }`: dragging in any direction stops at all three boundaries (e.g. dragging straight up from origin clamps to `(-3, 3, 3)` instead of escaping along Z).

### DIFF
--- a/examples/3d-scenes/draggable_constraints_3d.html
+++ b/examples/3d-scenes/draggable_constraints_3d.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>Draggable constraints in 3D (issue #260)</title>
+  <title>Draggable constraints in 3D (issues #260, #261)</title>
   <style>
     * { margin: 0; padding: 0; box-sizing: border-box; }
     body {
@@ -23,10 +23,10 @@
   </style>
 </head>
 <body>
-  <h1>Draggable Dot3D with constrainX / constrainY = [-3, 3] (issue #260)</h1>
+  <h1>Draggable Dot3D with constrainX / constrainY / constrainZ = [-3, 3] (issues #260, #261)</h1>
   <p>
-    Drag the dot. Try moving it past the [-3, 3] box: it must stop at the
-    boundary, not slide along the camera-facing plane in Z.
+    Drag the dot. Try moving it past the [-3, 3] cube on any axis: it must
+    stop at the boundary, not slide along the camera-facing plane.
   </p>
   <div class="pos" id="pos">position: (1.000, 1.000, 1.000)</div>
   <div id="container"></div>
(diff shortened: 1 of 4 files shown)


maloyan:
## Review feedback addressed

Ran a multi-agent review (code, tests, comments, silent failures). All four agents converged on one blocker — the JSDoc said `constrainZ` was "3D scenes only" but the implementation also clamped `mobject.position.z` in 2D scenes, silently mutating Z on the first drag tick. Fixed in c684cdc.

**Behavior change vs. previous push:**
- `constrainZ` clamp is now inside the `instanceof ThreeDScene` branch — 2D scenes are no longer touched.
- The 2D test was previously asserting the bug (`toBe(1)`) instead of the contract; now asserts `toBe(5)` (Z preserved).

**Test gaps closed:**
- `constrainZ overrides freeze-Z when frozen Z is outside range` — composes the issue #260 guard with the new clamp: starting at `[0, 0, 0.5]` with `constrainZ: [-0.4, 0.4]`, dragging until X clamps and Z freezes — asserts the clamp still pulls Z inside `[-0.4, 0.4]` instead of stalling at 0.5.
- `snapToGrid snaps Z in 3D scenes` — covers the previously untested 3D-only Z snap branch.

96/96 in the interaction suite, 5849/5849 full suite.

**Out of scope (filed mentally for follow-up):**
- Snap-after-clamp can re-violate the constraint range (e.g. `constrainZ: [-2, 2]` + `snapToGrid: 3` → snapped to 3). This pattern pre-dates the PR and applies to X/Y identically; fixing it is a separate change.
- `[NaN, NaN]` bounds aren't validated — same pre-existing pattern.
- Initial-position-out-of-bounds doesn't get clamped at `_startDrag`. Edge case, separate change.

PR #307, opened 2026-05-01
fix: Transform for ImageMobject and Text with geometry morphing

### LINKED ISSUE #305: When text is scaled up, the scene ends with original size
## Description                                                                                                                                                                          
Text scale animation reverts to original size after the animation completes.                                                                                                            
                                                                                                                                                                                       
## Steps to Reproduce                                                                                                                                                                   
1. Create a Text mobject                                                                                                                                                                
2. Play FadeIn animation                                                                                                                                                                
3. Play scale(2) animation via animate                                                                                                                                                  
```js
import { Scene, Text, FadeIn } from "https://cdn.jsdelivr.net/npm/manim-web@latest/dist/manim-web.browser.js";                                                                          
                                                                                                                                                                                       
const scene = new Scene(document.getElementById("container"), { width: 500, height: 300 });                                                                                             
const text = new Text({ text: "Hello World" });                                                                                                                                         
                                                                                                                                                                                       
await scene.play(new FadeIn(text));                                                                                                                                                     
await scene.play(text.animate.scale(2));                                                                                                                                                
```

## Expected Behavior                                                                                                                                                                    
Text should remain scaled 2x after the animation completes.                                                                                                                             
                                                                                                                                                                                       
## Actual Behavior                                                                                                                                                                      
Text scales up during animation, then immediately reverts to original size when animation ends.                                                                                         
                                                                                                                                                                                       
## Environment                                                                                                                                                                          
- Browser: firefox                                                                                                                                                    
- OS: linux                                                                                                                                                                    
- manim-web version: latest (via CDN)

### LINKED ISSUE #306: ImageMobject MoveToTarget does not apply scale transformation
## Description                                                                                                                                                                          
ImageMobject scale animation via MoveToTarget does not work correctly - the image does not scale when using MoveToTarget, while direct scale() works.                                   
                                                                                                                                                                                       
## Steps to Reproduce                                                                                                                                                                   
1. Create an ImageMobject                                                                                                                                                               
2. Call generateTarget() and target.scale(2)                                                                                                                                            
3. Play MoveToTarget animation                                                                                                                                                          
```js
import { Scene, ImageMobject, FadeIn, MoveToTarget } from "https://cdn.jsdelivr.net/npm/manim-web@latest/dist/manim-web.browser.js";                                                    
                                                                                                                                                                                       
const scene = new Scene(document.getElementById("container"), { width: 500, height: 300 });                                                                                             
const image = new ImageMobject({ source: "...", height: 2 });                                                                                                                           
                                                                                                                                                                                       
await scene.play(new FadeIn(image));                                                                                                                                                    
image.generateTarget();                                                                                                                                                                 
image.target.scale(2);                                                                                                                                                                  
await scene.play(new MoveToTarget(image));                                                                                                                                              
```
## Expected Behavior                                                                                                                                                                    
Image should scale up to 2x and maintain that size after animation.                                                                                                                     
                                                                                                                                                                                       
## Actual Behavior                                                                                                                                                                      
Image remains at original size - MoveToTarget does not apply the transformation.                                                                                                        
                                                                                                                                                                                       
## Environment                                                                                                                                                                          
- Browser: (fill in)                                                                                                                                                                    
- OS: (fill in)                                                                                                                                                                         
- manim-web version: latest (via CDN)

rambip:
## Summary

Fix and improved Transform animation for ImageMobject and Text mobjects.

Now image and text transforms are way more powerful:

https://github.com/user-attachments/assets/f44c828b-adf2-4634-88a8-8bb8c3292416

https://github.com/user-attachments/assets/e7b45871-28f1-48cd-8cd6-a62c942e549e

Previously, ImageMobject scale via MoveToTarget failed silently, and Text scale reverted to original size after animation.

## Changes
- Add `MorphMode` enum: `Point` (VMobject points), `Fade` (opacity+position), `Shape` (geometry+transform)
- Shape mode: cross-fade + geometry dimensions + position/rotation/scale for ImageMobject/Text/MathTexImage
- Fade mode: cross-fade opacity + position only for zero-point VMobjects and non-VMobjects  
- Point mode: standard VMobject point interpolation
- Add helper methods: `_isShapeModeType()`, `_captureTransformProps()`, `_captureGeometryDimensions()`, `_setupCrossFade()`, `_swapTexture()`, `_interpolatePosition()`, `_interpolateTransform()`, `_interpolateCrossFade()`
- Add `hasSameContentAs()` to Mobject base class
- Fix: check Shape mode types with `instanceof` before VMobject check (ImageMobject extends Mobject, not VMobject)

## Testing
- [x] Tests pass (`bun test`) 
- [x] Lint passes (`bun run lint`)

## Related Issues
Closes #305 - Text scale now preserves final size
Closes #306 - ImageMobject MoveToTarget now applies scale transformation


rambip:
@maloyan I wait for your review !

src/animation/transform/Transform.ts
```diff
@@ -128,39 +153,140 @@ export class Transform extends Animation {
...
+  private _captureGeometryDimensions(): void {
+    const srcMesh = this.mobject.getThreeObject() as THREE.Mesh;
+    const tgtMesh = this.target.getThreeObject() as THREE.Mesh;
+    const srcGeo = srcMesh.geometry as THREE.PlaneGeometry;
+    const tgtGeo = tgtMesh.geometry as THREE.PlaneGeometry;
+    this._startWidth = srcGeo.parameters.width;
+    this._startHeight = srcGeo.parameters.height;
+    this._targetWidth = tgtGeo.parameters.width;
+    this._targetHeight = tgtGeo.parameters.height;
+  }
```
maloyan: **Critical — runtime crash on Text→Text and MathTexImage→MathTexImage.**

`Text._createThreeObject()` returns `THREE.Group` containing the mesh; `MathTexImage._createThreeObject()` does the same (and returns an *empty* group when `_isMultiPart=true`). Only `ImageMobject` returns a `Mesh` directly.

The `as THREE.Mesh` cast is TypeScript-only — at runtime `srcMesh.geometry` is `undefined` on a Group, so `srcGeo.parameters.width` throws `TypeError: Cannot read properties of undefined (reading 'parameters')`.

This is the very path the PR claims to support (closes #305 — Text scale). CI is green because no test exercises real `Text`/`MathTexImage` Transforms; existing tests use a `TextLikeMobject` stand-in that fails the `instanceof Text` guard and silently routes through Fade.

Same unsafe cast appears at `interpolate()` (~L425), `finish()` (~L546), and `_swapTexture` (~L197).

**Fix:** add a typed accessor like `getDisplayMesh(): THREE.Mesh` on the Shape-mode classes (or unwrap from Group), and validate `geometry instanceof PlaneGeometry` before reading `.parameters`. Multi-part MathTex (empty Group) should fall back to Fade with a warn.

src/animation/transform/Transform.ts
```diff
@@ -128,39 +153,140 @@ export class Transform extends Animation {
...
+    const targetMesh = this.target.getThreeObject() as THREE.Mesh;
+    const srcMat = sourceMesh.material as THREE.MeshBasicMaterial;
+    const tgtMat = targetMesh.material as THREE.MeshBasicMaterial;
+    srcMat.map = tgtMat.map;
+    srcMat.needsUpdate = true;
+    const src = this.mobject as TexturedMobject;
+    const tgt = this.target as TexturedMobject;
+    src._source = tgt._source;
+    src._texture = tgtMat.map;
+  }
```
maloyan: **Critical — corrupts MathTexImage state and leaks GPU textures.**

Three distinct problems:

1. **MathTexImage state desync.** `MathTexImage._syncMaterialToThree()` re-reads from `_renderState.texture/width/height` on the next dirty cascade and reverts both `material.map` and the geometry. Since `_swapTexture` never updates `_renderState`, any subsequent style change, opacity tween, or `_markDirty` reverts the swap. The Shape→MathTexImage path is unstable past `finish()`.

2. **Phantom `_source` field on Text/MathTexImage.** `_source` exists only on `ImageMobject` (`src/mobjects/image/index.ts:76`). `Text` (`Text.ts:127`) has `_texture` only. `MathTexImage` has neither — uses `_renderState.texture`. The `TexturedMobject` interface is a fiction for two of three classes; the assignment writes `undefined` into a phantom property name.

3. **GPU texture leak.** `srcMat.map = tgtMat.map` and `src._texture = tgtMat.map` overwrite without `.dispose()`-ing the prior `THREE.Texture`. Each Shape Transform leaks a GPU texture. Ownership also becomes ambiguous — both source and target now reference the same GPU resource.

**Fix:** type-specific swap helpers (`_swapImageMobject(src, tgt)`, `_swapText(src, tgt)`, `_swapMathTexImage(src, tgt)`), each updating the class's authoritative state (including `_renderState` for MathTexImage) and disposing the displaced texture.

src/animation/transform/Transform.ts
```diff
@@ -128,39 +153,140 @@ export class Transform extends Animation {
...
    */
   override begin(): void {
     super.begin();
 
+    // Shape mode: ImageMobject/Text/MathTexImage with geometry morphing
+    // Check first since ImageMobject is NOT a VMobject
+    if (
+      this._isShapeModeType(this.mobject) &&
+      this._isShapeModeType(this.target) &&
+      this.mobject.constructor === this.target.constructor
```
maloyan: **Important — drops legitimate Shape pairs into Fade.**

`this.mobject.constructor === this.target.constructor` is exact-class identity. `Paragraph extends Text` and `MarkupText extends Text` (`Paragraph.ts:38`, `MarkupText.ts:633`), so `Transform(text, paragraph)` fails the equality, falls through to the VMobject branch, fails `canMorphByPoints`, and lands in Fade — which then makes the result invisible (see L566 comment).

**Fix:** prefer capability-based check: `m1 instanceof Text && m2 instanceof Text` (and similarly for the other two classes), or define a shared marker interface.

src/animation/transform/Transform.ts
```diff
@@ -128,39 +153,140 @@ export class Transform extends Animation {
...
+        vmobject.constructor === vtarget.constructor
+      ) {
+        this._morphMode = MorphMode.Shape;
         this._startOpacity = this.mobject.opacity;
-        this._crossFadeTargetOpacity = this.target.opacity;
+        this._captureTransformProps();
+        this._captureGeometryDimensions();
+        this._setupCrossFade();
+        return;
+      }
```
maloyan: **Important — dead code.**

This branch is unreachable. The top-level Shape-mode check at L249 fires for any `_isShapeModeType` pair regardless of VMobject inheritance and `return`s. So by the time control reaches L269, at least one of the two is not a Shape type — meaning the inner `_isShapeModeType && _isShapeModeType` test can never be true.

Verified: `Text extends VMobject`, `ImageMobject extends Mobject`, `MathTexImage extends Mobject`. Both-Shape-type pairs always hit the top check.

**Fix:** delete L269-281.

src/animation/transform/Transform.ts
```diff
@@ -312,17 +418,30 @@ export class Transform extends Animation {
...
+      const w = this._startWidth + (this._targetWidth - this._startWidth) * alpha;
+      const h = this._startHeight + (this._targetHeight - this._startHeight) * alpha;
+
+      const sourceMesh = this.mobject.getThreeObject() as THREE.Mesh;
+      sourceMesh.geometry.dispose();
+      sourceMesh.geometry = new THREE.PlaneGeometry(w, h);
+
+      const targetMesh = this.target.getThreeObject() as THREE.Mesh;
+      targetMesh.geometry.dispose();
+      targetMesh.geometry = new THREE.PlaneGeometry(w, h);
```
maloyan: **Important — per-frame GPU buffer churn.**

Every `interpolate()` tick disposes and constructs two `THREE.PlaneGeometry` instances. At 60fps that's 120 GPU buffer allocations per second per Shape transform. `PlaneGeometry` allocates BufferAttributes (position, normal, uv, index) — meaningful GC pressure on lower-end devices.

**Cheaper option:** scale a unit plane instead — `mesh.scale.set(w, h, 1)`. Visible result for an axis-aligned plane is identical. Build geometry once in `begin()`, set final dims in `finish()`. Note: you'd need to separate display-scale from `mobject.scaleVector` since the latter is also being interpolated.

src/animation/transform/Transform.ts
```diff
@@ -422,56 +541,39 @@ export class Transform extends Animation {
...
+      this.target.opacity = this._crossFadeTargetOpacity;
+      this.target._syncToThree();
+
+      const targetObj = this.target.getThreeObject();
+      if (targetObj.parent) targetObj.parent.remove(targetObj);
+
+      this.mobject._markDirty();
       super.finish();
       return;
     }
```
maloyan: **Critical — Fade-mode finish leaves nothing visible.**

```
this.mobject.opacity = 0;          // source invisible
...
targetObj.parent.remove(targetObj); // target detached
```

Net result: source still in scene with opacity 0, target removed from scene graph. Nothing renders.

The pre-PR `finish()` (deleted in this diff) reparented `targetObj` under `sourceObj` after computing world-position deltas, so target stayed visible attached to source. That code path is gone, but the situation it handled (zero-point VMobject pairs and non-VM/non-Shape pairs) still routes here.

Downstream effect: any chained animation on `mobject` (e.g. `FadeOut(mobject)`) sees an invisible source with stale state and is a no-op. User sees "transform happened, then everything disappeared."

**Fix:** either restore the reparent-target-under-source step from the old code, or copy target's visual state (texture/material/geometry) onto source à la Shape mode and set `mobject.opacity = _crossFadeTargetOpacity`.

src/core/Mobject.ts
```diff
@@ -641,6 +641,13 @@ export abstract class Mobject {
     return restoreMobjectStateImpl(this);
   }
 
+  // ── Comparison ───────────────────────────────────────────────────
+
+  /** Check if this mobject has the same content as another. Default: false. */
+  hasSameContentAs(_other: Mobject): boolean {
```
maloyan: **Suggestion — dead method.**

`grep -rn hasSameContentAs src/` returns only this definition. No callers. The doc says "Default: false" without explaining when subclasses should override or which consumer reads it.

**Options:**
- Drop until a consumer needs it (preferred — ship API surface with its caller).
- If kept, document the contract: which subclasses override, what the caller (Transform pairing? animation scheduling?) does with `true`.

src/animation/transform/Transform.ts
```diff
@@ -18,6 +18,25 @@ import {
   pairLeafSnapshotsByIndex,
   type LeafPairByIndex,
 } from './TransformPairing';
+import { ImageMobject } from '../../mobjects/image';
+import { Text } from '../../mobjects/text/Text';
```
maloyan: **Suggestion — file is now 699 lines (was 497), 40% over the `max-lines: 500` ESLint soft limit.**

The file now does four distinct algorithms: Point morphing, VGroup per-child morphing, Fade cross-fade, Shape geometry morph. Worth extracting Shape and VGroup paths into helper modules (e.g. `TransformShape.ts`, `TransformVGroup.ts`) so this file stays readable.

Not a blocker (the rule is `warn`), but the divergence keeps growing.

examples/image_scale_direct.html
```diff
@@ -0,0 +1,59 @@
...
+        body { font-family: system-ui, sans-serif; padding: 20px; }
+        .container { border: 1px solid #ccc; display: inline-block; margin: 10px; }
+    </style>
+</head>
+<body>
+    <h2>Image Scale Direct Test</h2>
+    <p>Left: original (height=2). Right: scaled by 2.</p>
+    <div id="container1" class="container"></div>
+    <div id="container2" class="container"></div>
+
```
maloyan: **Nit — trailing whitespace.** `git diff --check origin/main...HEAD` flags trailing whitespace here and in `examples/image_transform.html:78` and `examples/widget_logo_test.html:58`. Strip and re-push.

maloyan (changes requested):
Comprehensive review (4 specialized agents + Codex independent pass). All converged on the same critical bugs; flagging on the lines below.

**Verdict: block merge.** The Shape-mode path is broken for Text and MathTexImage (the very types this PR claims to support — see #305), Fade-mode finish leaves the scene empty, and `_swapTexture` desyncs MathTexImage internal state. CI is green only because no test exercises the new path; existing "Text-like" tests use stubs that bypass the new `instanceof Text` guard.

**Must-fix before merge:**
1. Extract inner Mesh from Group for Text/MathTexImage; guard multi-part MathTex.
2. Per-class swap helpers; update `MathTexImage._renderState`; dispose old textures.
3. Restore Fade-mode reparent (or copy target visual onto source).
4. Loosen constructor-equality guard so subclass relationships work.
5. Broaden `MobjectWithTarget` (Transform.ts:679) and `MoveToTarget` to accept `Mobject` not `VMobject` — currently `ImageMobject extends Mobject`, so the #306 fix is unreachable from typed code without a cast.
6. Add regression tests for #305 and #306 — none exist.
7. Remove dead block; drop unused `hasSameContentAs` or wire it up.

**Nice-to-fix:**
- Per-frame `PlaneGeometry` allocation (use mesh.scale instead).
- Trailing whitespace in example HTMLs (`git diff --check` fails).
- Strip JSDoc that just restates signatures.
- `bun.lock` (+1619 lines) noise — confirm intentional.

Positives: `MorphMode` enum cleaner than the old boolean; helper extraction reduces duplication; Shape-mode early-return correctly precedes the VMobject branch.

rambip:
@maloyan ready for new review.
I tried to follow your advice

I'm not sure where this file comes from:
https://github.com/maloyan/manim-web/blob/56254eb075e32f2ef50eb3c3e934a5d49974d7af/examples/mathtex_stroke_zero.html

Also, I'm not sure about `TexturedMobject` for `Groups`: I think it should not do a `ShapeMorph` in this case but not sure

src/animation/transform/FadeMorphStrategy.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (!this._targetSnapshot) {
+      throw new Error('FadeMorphStrategy.finish requires target snapshot');
+    }
+    source.position.set(this._targetPositionX, this._targetPositionY, this._targetPositionZ);
+    source.opacity = 0;
+    target.position.set(this._targetPositionX, this._targetPositionY, this._targetPositionZ);
+    target.opacity = this._targetSnapshot.opacity;
+    target._syncToThree();
+    const targetObj = target.getThreeObject();
+    if (targetObj.parent) targetObj.parent.remove(targetObj);
```
maloyan: **Important — fade-mode `finish()` leaves nothing visible.**

```
source.opacity = 0;                  // line 73 — source invisible
...
targetObj.parent.remove(targetObj);  // line 78 — target detached from scene graph
```

Net result: source still in scene with opacity 0, target removed. Nothing renders. Any chained animation on `source` (e.g. `FadeOut(source)`) sees an invisible mobject and is a no-op. User sees "transform happened, then everything disappeared."

This is the same regression pattern flagged in an earlier resolved comment on `Transform.ts` and survived the strategy extraction.

**Fix:** either copy target visual state onto source (à la ShapeMorph) and set `source.opacity = _targetSnapshot.opacity`, or restore the pre-PR reparent-target-under-source step so the target stays visible attached to source.

src/animation/transform/ShapeMorphStrategy.ts
```diff
@@ -0,0 +1,156 @@
...
+    target.position.copy(source.position);
+    source.rotation.set(
+      this._startRotation.x + (this._targetRotation.x - this._startRotation.x) * alpha,
+      this._startRotation.y + (this._targetRotation.y - this._startRotation.y) * alpha,
+      this._startRotation.z + (this._targetRotation.z - this._startRotation.z) * alpha,
+    );
+    target.rotation.copy(source.rotation);
+    source.scaleVector.lerpVectors(this._startScale, this._targetScale, alpha);
+    target.scaleVector.copy(source.scaleVector);
+    source.opacity = this._crossFadeTargetOpacity * (1 - alpha);
```
maloyan: **Important — opacity pop on first frame.**

`_crossFadeTargetOpacity` snapshots only the **target's** opacity in `begin()` (`this._crossFadeTargetOpacity = target.opacity`). At α=0 this line evaluates to `source.opacity = targetOpacity * 1 = targetOpacity`.

If source was opacity `1.0` and target opacity `0.5`, source jumps `1.0 → 0.5` on the very first interpolate tick — visible pop.

**Fix:** capture source opacity separately in `begin()`:
```ts
this._crossFadeStartOpacity = source.opacity;
```
then lerp:
```ts
source.opacity = this._crossFadeStartOpacity * (1 - alpha);
target.opacity = this._crossFadeTargetOpacity * alpha;
```

src/animation/transform/ShapeMorphStrategy.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  finish(_animation: Animation, source: Mobject, _target: Mobject): void {
+    if (!this._shapeSourceMesh) throw new Error('ShapeMorphStrategy requires _shapeSourceMesh');
+    if (!this._shapeTargetMesh) throw new Error('ShapeMorphStrategy requires _shapeTargetMesh');
+    if (!this._sourceTextureMobject)
+      throw new Error('ShapeMorphStrategy requires source TexturedMobject');
+    if (!this._targetTextureMobject)
+      throw new Error('ShapeMorphStrategy requires target TexturedMobject');
+
+    this._sourceTextureMobject.applyVisualSize(this._targetWidth, this._targetHeight);
```
maloyan, line 150: **Important — content metadata stays stale after Shape morph.**

`applyVisualSize` + `applyTextureFrom` transfer the *visual* state, but the authoritative content fields on the concrete classes are never updated:
- `Text.text` (see `src/mobjects/text/Text.ts:529`)
- `ImageMobject` source URL (see `src/mobjects/image/index.ts:462`)
- `MathTexImage` tex string + `_renderState` (see `src/mobjects/text/MathTexImage.ts:1038`)

After `Transform(Text("A"), Text("B")).finish()`:
- `source.getText()` still returns `"A"`
- A later `_markDirty()` / `_syncToThree()` re-renders from `"A"` and reverts the texture
- `source.copy()` produces a `Text("A")` with B's texture — silently inconsistent

Same failure mode as the resolved `_swapTexture` desync comment, just relocated to the strategy.

**Fix:** add a typed `applyContentFrom(target: this)` on `TexturedMobject` (or per-class `_swapText`/`_swapImage`/`_swapMathTex` helpers) that copies the authoritative content state in addition to the texture, and dispose the previously-owned texture to avoid GPU leaks.

src/animation/transform/Transform.ts
```diff
@@ -539,59 +73,26 @@ export function transform(
...
   options?: AnimationOptions,
 ): ReplacementTransform {
   return new ReplacementTransform(mobject, target, options);
 }
-
-/**
- * MoveToTarget animation - moves a mobject to its target copy.
- * The mobject must have a `.targetCopy` property set beforehand.
- */
 export interface MobjectWithTarget extends VMobject {
```
maloyan: **Important — `MoveToTarget` typing blocks the very classes this PR fixes.**

`MobjectWithTarget extends VMobject` with `targetCopy: VMobject | null`, but:
- `targetCopy` lives on `Mobject` base (`src/core/Mobject.ts:93`)
- The PR's headline fix (#306) is `ImageMobject.MoveToTarget` — and `ImageMobject extends Mobject`, not `VMobject`
- Same for `Text` after the new `extends TexturedMobject` change

Result: callers must cast through `as unknown as MobjectWithTarget` to use the very path the PR claims to enable. CI passes only because the new test constructs the path manually.

**Fix:**
```ts
export interface MobjectWithTarget extends Mobject {
  targetCopy: Mobject | null;
}
```
(Or drop the interface and inline `Mobject & { targetCopy: Mobject | null }` at the constructor.)

src/animation/transform/PointMorphStrategy.ts
```diff
@@ -0,0 +1,268 @@
...
+  return {
+    opacity: 0,
+    fillOpacity: 0,
+    strokeWidth: m.strokeWidth,
+    color: m.color,
+    fillColor: m.fillColor || m.color,
+  };
+}
+
+export class PointMorphStrategy {
```
maloyan: **Suggestion — missing `implements MorphStrategy`.**

`FadeMorphStrategy` and `ShapeMorphStrategy` declare `implements MorphStrategy`; this class relies on structural typing alone. If the interface gains/changes a method, that drift won't surface here at compile time.

```ts
export class PointMorphStrategy implements MorphStrategy {
```

src/animation/transform.test.ts
```diff
@@ -423,6 +425,60 @@ describe('MoveToTarget', () => {
...
+
+      const move = new MoveToTarget(image as unknown as MobjectWithTarget);
+      move.begin();
+      move.interpolate(1);
+      move.finish();
+
+      const finalWidth = image.getBoundingBox().width;
+      expect(finalWidth).toBeCloseTo(expectedWidth, 5);
+      expect(finalWidth).toBeGreaterThan(startWidth);
+    } finally {
```
maloyan, line 479: **Important — #305 regression coverage missing.**

The PR claims to close #305 (Text scale revert) but only adds a regression test for #306 (ImageMobject scale via MoveToTarget). The original #305 bug was that *a later sync cycle* reverted the size set by Transform finish.

Minimal addition:
```ts
it('preserves Text scale after Transform finish and subsequent _syncToThree', () => {
  const t1 = new Text('A');
  const t2 = new Text('B').scale(2);
  const tr = new Transform(t1, t2);
  tr.begin(); tr.interpolate(0.5); tr.finish();
  const w1 = t1.getBoundingBox().width;
  t1._markDirty(); t1._syncToThree();
  expect(t1.getBoundingBox().width).toBeCloseTo(w1, 5);
});
```
Without this, the very bug being closed can silently regress on the next render-pipeline change.

maloyan (commented):
Multi-agent review (codex + reviewer agents) reconciled with author's earlier resolved comments. Inline issues below. Several historical concerns are correctly addressed at HEAD (Group/Mesh cast via `getDisplayMeshes()`, per-frame PlaneGeometry alloc via unit plane + mesh.scale, file size via Strategy split, `constructor === constructor` replaced with `instanceof`). Remaining items are post-refactor regressions or freshly introduced gaps.

**Action plan**
1. Fix Fade `finish()` visibility regression.
2. Capture source opacity in ShapeMorph (currently pops at α=0).
3. Sync content metadata in Shape `finish()` so `getText()`/source URL/tex string reflect target.
4. Widen `MobjectWithTarget` so the ImageMobject/Text path the PR claims to fix is callable without casts.
5. Add #305 regression test for Text scale persistence after `_syncToThree()`.
6. Add `implements MorphStrategy` to `PointMorphStrategy`.
7. Update PR description — `MorphMode`, `hasSameContentAs`, `_swapTexture`, `_captureGeometryDimensions`, `_setupCrossFade`, `_interpolateCrossFade` are not in the diff at HEAD.

rambip:
@copilot read review, that seems like pretty simple to fix. Prefer assertions over fallbacks. you have 6 unresolved items to fix.

rambip:
@maloyan should be good to merge now !

maloyan (approved):

PR #319, opened 2026-05-05
Fix/transform compound alignment

rambip:
## Summary

#298 Added some improvements regarding the endstate of SVG math TeX, but the interpolation still had quite a lot of issues.

This pull request greatly improves the initial alignments of points and the interpolation step. It chooses a logic that matches a bit more Manim Python regarding resampling the curves. It also corrects a bug found in SVG parsing.

**Before**:  😕

https://github.com/user-attachments/assets/44438be3-3a48-48a9-973f-682dd0c956fc

**After**: 😊

https://github.com/user-attachments/assets/28aaf86b-887a-4759-822f-d94a15bccb85

## Changes
- Extracted transform alignment into src/core/VMobjectTransformAlignment.ts; VMobjectGeometry.ts now re-exports alignment API.                                                           
 - Improved compound morph pairing: orientation-first matching (CW/CCW buckets), then length sorting.                                                                                     
 - Added closed-subpath target anchor rotation to align first anchors with source (better first-frame continuity).                                                                        
 - Wired orientation-aware alignment through transform matching/pairing callsites.                                                                                                        
 - Added/updated regressions in:                                                                                                                                                          
     - src/animation/transform/TransformPairing.test.ts (orientation pairing + anchor rotation)                                                                                           
     - src/animation/transform.test.ts (MathTex 1 -> 0 first-frame point closeness)                                                                                                       
     - src/animation/animation-transform-extra.test.ts (metadata expectation to preserved [5,5], not [64,64])                                                                             
 - Verified: targeted transform tests pass, lint/typecheck clean.        

## Testing
- [x] Tests pass (`npm test`)


maloyan (commented):
## Multi-agent review summary (5 specialised agents + Codex independent verification)

Reviewed 13 changed files (+984/-220). Tests pass, lint clean. Found one likely behavioural regression outside the stated scope (compound fill triangulation), plus several silent-failure / metadata-validation gaps in the new alignment module. Inline comments below.

### Highlights

**High** — Compound fill triangulation rewrite (`VMobjectGeometry.ts`) picks a single longest-perimeter ring as primary outer and renders all other outer rings independently *without their holes*. Disjoint outer contours that each have holes (e.g. `%`, two-disjoint-donut shapes, multi-glyph fills) lose hole subtraction. This change is outside the PR's stated scope ("transform compound alignment") and has no regression test.

**Medium** — Several silent-failure paths in the new alignment module:
- Zero/near-zero `signedArea2D` collapses to CCW (+1) — degenerate or thin glyphs (`-`, `=`, `i`-dot) silently bucket wrong.
- `(srcLengths.length <= 1 && tgtLengths.length <= 1)` returns null *before* `validateSubpathLengths`. With the SVG parser fix attaching `[N]` for single-contour glyphs, malformed metadata can now skip validation.
- `resamplePointList3D` fabricates `[0, 0, 0]` origin points on empty input — fights the upstream `makeNullSubpathFromReference` fallback.
- `getSubpathOrientationSigns` returns `undefined` on inconsistent metadata with no log/throw; caller silently falls back to per-chunk inference.
- `sameSignPairs === 0` fallback length-pairs all-CW src with all-CCW tgt without reversing winding — mirrored-glyph morphs may twist mid-interpolation.

**Low** — Test/diagnostic hygiene:
- `transform.test.ts:142-144` empty catch swallows the very `validateSubpathLengths`/`alignVmobjectPair` invariants this regression test was added to enforce.
- `getSubpathOrientationSign(index)` falls back to whole-shape area on out-of-range — masks caller bugs.
- `signedArea2D` duplicated in `VMobject._signedArea2D` (line 445) and `VMobjectTransformAlignment.ts` (line 288); two oracles, no test pins equivalence.
- `VMobjectGeometry.ts` re-exports the alignment API only as a compatibility facade; direct imports would clarify ownership.
- `resamplePointList3D` JSDoc says "linear interpolation" / "mirrors `_interpolatePointList3D`" but the function actively prefers cubic-Bezier subdivision.

### Test coverage gaps
- No direct unit tests for new module internals (`resamplePointList3D` cubic↔linear branch boundaries, `subdivideCubicCurve`, `remapBezierCurveCount`, `signedArea2D`).
- `getSubpathOrientationSigns` / `getSubpathOrientationSign` (new public API) untested directly.
- All-CW vs all-CCW (`sameSignPairs === 0`) fallback path completely uncovered.
- Auto-inference of signs (omitting the new `srcSigns`/`tgtSigns` params) never exercised — every test passes explicit signs.

### Strengths
- SVG parser fix `> 1` → `> 0` is correct and pinned by direct regression.
- `[64, 64] → [5, 5]` metadata change correctly reflects the removed `MIN_TRANSFORM_SUBPATH_POINTS` floor.
- `Array<1 | -1>` literal type is a good encoding for the orientation invariant.
- Closed-target anchor rotation logic is sound (stride preserved, dedicated regressions).
- Orientation-aware wiring is consistent across all three call sites.
- New module respects the 500-line ESLint cap (420 lines).
- No new `any` / `unknown` introduced.

### Recommended action
1. Split the fill-triangulation rewrite into a separate PR with its own multi-outer-with-holes regression — or revert that block here.
2. Address the metadata-validation / silent-CCW classification issues (the alignment module's primary correctness surface).
3. Remove the empty catch in the MathTex regression test.
4. Add direct unit tests for the new alignment helpers and the omitted-signs auto-inference path.

src/core/VMobjectGeometry.ts
```diff
@@ -640,58 +472,67 @@ function buildEarcutFillGeometryMulti(
...
-    for (let j = 0; j < rings2D.length; j++) {
-      if (i === j) continue;
-      if (pointInPolygon(rings2D[i][0], rings2D[j])) {
-        isHoleOf[i] = j;
-        break;
-      }
+  // Heuristic classification: pick longest ring as primary outer contour.
+  // Treat only rings INSIDE that contour as holes; rings outside stay
+  // independent outers (prevents '=' turning into one bar).
+  let primaryOuterRingIndex = 0;
```
maloyan: **High — likely regression outside PR scope.**

This heuristic picks the *single* longest-perimeter ring as `primaryOuterRingIndex`, then below treats only rings inside that primary as holes; every other outer ring is pushed to `independentOuterRingIndices` and triangulated *without* its hole list.

For any compound shape with two or more disjoint outer contours where a non-primary outer has holes — e.g. `%`, two stacked annuli, multi-glyph SVG fills — those holes will be filled instead of subtracted. The old code computed full ring-containment so every outer carried its own holes.

The PR title and description are about transform alignment; this fill-rendering rewrite has no test and no mention. Either:

1. Replace with a parity / smallest-enclosing-ring containment hierarchy and triangulate each outer with its direct hole children, **and** add a regression with two disjoint outers each containing a hole; or
2. Split this change into a separate PR.

The `VMobjectGeometry.test.ts` case at `:83-126` only exercises one outer + one hole, so it does not catch this.

src/core/VMobjectGeometry.ts
```diff
@@ -19,12 +19,8 @@ import {
   projectToPlane,
 } from '../utils/math';
 import type { Point } from './VMobjectCurves';
-
-export interface CompoundAlignmentResult {
-  srcAlignedPoints: number[][];
-  tgtAlignedPoints: number[][];
-  alignedSubpathLengths: number[];
-}
+export { alignCompoundPathsForTransform } from './VMobjectTransformAlignment';
```
maloyan: **Low — leaky re-export.**

Every existing importer (`TransformMatching.ts`, `TransformMatchingBase.ts`, `TransformPairing.ts`, `TransformPairing.test.ts`) still pulls `alignCompoundPathsForTransform` from `./VMobjectGeometry` even though the symbol now lives in `./VMobjectTransformAlignment`. This re-export only exists to avoid touching call sites and keeps `VMobjectGeometry` coupled to alignment code it no longer owns.

Single-pass fix: update the four importers to import directly from `./VMobjectTransformAlignment` and drop these two lines.

src/core/VMobjectTransformAlignment.ts
```diff
@@ -0,0 +1,420 @@
...
+}
+
+/**
+ * Resample a 3D point list to a target count using linear interpolation.
+ *
+ * This mirrors VMobject._interpolatePointList3D behavior and is used for
+ * compound-path pairing so the shorter subpath is resampled to match the
+ * longer subpath's point count.
+ */
+function resamplePointList3D(points: number[][], targetCount: number): number[][] {
```
maloyan, line 37: **Low — JSDoc inaccurate.**

The doc says "linear interpolation" and "mirrors `VMobject._interpolatePointList3D`", but the function takes a Bezier-aware fast path (`isCubicBezierChain` + `remapBezierCurveCount`) and only falls back to lerp for non-cubic / mismatched-stride cases. `_interpolatePointList3D` is pure lerp — these are not mirrors.

Suggest: *"Resample a 3D point list to `targetCount`. For cubic-Bezier-chain inputs whose target count fits the cubic stride (`(n-1) % 3 === 0`), uses De Casteljau-style subdivision via `remapBezierCurveCount` to preserve curve fidelity; otherwise falls back to piecewise-linear interpolation."*

src/core/VMobjectTransformAlignment.ts
```diff
@@ -0,0 +1,420 @@
...
+  if (targetCount <= 0) return [];
+
+  if (points.length === 0) {
+    return Array.from({ length: targetCount }, () => [0, 0, 0]);
```
maloyan: **Medium — fabricates world-origin points on empty input.**

When the input chunk is empty, the function silently returns `targetCount` copies of `[0, 0, 0]`. The legitimate "missing chunk" path is `makeNullSubpathFromReference` (which anchors at the prior subpath endpoint); this lower-level fallback fights the higher-level one and pins a phantom subpath at world centre, producing a visible morph artifact at `(0,0,0)`.

Fix: `throw` here. If a zero-length chunk reaches `resamplePointList3D`, that's a metadata bug upstream and should be loud.

src/core/VMobjectTransformAlignment.ts
```diff
@@ -0,0 +1,420 @@
...
+interface SignedSubpathChunk {
+  chunk: number[][];
+  len: number;
+  sign: 1 | -1;
+}
+
+function signedArea2D(pts: number[][]): number {
+  const anchors: number[][] = [];
+  for (let i = 0; i < pts.length; i += 3) anchors.push(pts[i]);
+  if (anchors.length < 3) return 0;
```
maloyan: **Medium — degenerate / near-zero area silently classified as CCW.**

`signedArea2D` returns `0` for `< 3` anchors / collinear / genuinely degenerate input. Combined with `inferSign` below (`area < 0 ? -1 : 1`), every such chunk is bucketed as CCW (+1) without warning. Thin glyph fragments from LaTeX (`-`, `=`, dotless parts of `i`/`j`) hit this. Float-noise near zero is also non-deterministic across runs, flipping a contour between CW/CCW between source and target of the same identity transform.

Suggested: introduce an epsilon-based tri-state (or filter neutral chunks out of orientation buckets and pair them by length only). The same fix needs to land in `VMobject.getSubpathOrientationSigns` (line 333) — the two oracles must agree.

src/core/VMobjectTransformAlignment.ts
```diff
@@ -0,0 +1,420 @@
...
+ */
+export function alignCompoundPathsForTransform(
+  srcPoints3D: number[][],
+  srcLengths: number[] | undefined,
+  tgtPoints3D: number[][],
+  tgtLengths: number[] | undefined,
+  srcSigns?: Array<1 | -1>,
+  tgtSigns?: Array<1 | -1>,
+): CompoundAlignmentResult | null {
+  if (!srcLengths || !tgtLengths || (srcLengths.length <= 1 && tgtLengths.length <= 1)) {
```
maloyan, line 362: **Medium — early-return bypasses validation.**

`validateSubpathLengths` runs only on lines 358-359, *after* this early-return. With the SVG parser fix attaching `[N]` for single-contour glyphs (`svgPathParser.ts:637`), every single-glyph VMobject now hits this branch and skips the sum check. If lengths and points get desynchronised (e.g. an upstream `setPoints3D` that doesn't update lengths), the corruption silently propagates to `alignPoints` fallback and then to other code paths.

Fix: validate first, *then* early-return.

```ts
if (srcLengths) validateSubpathLengths(srcPoints3D, srcLengths, 'source');
if (tgtLengths) validateSubpathLengths(tgtPoints3D, tgtLengths, 'target');
if (!srcLengths || !tgtLengths || (srcLengths.length <= 1 && tgtLengths.length <= 1)) return null;
```

Also missing: `srcSigns.length === srcLengths.length` parity check — mismatched arrays silently fall through to `inferSign` for missing indices.

src/core/VMobjectTransformAlignment.ts
```diff
@@ -0,0 +1,420 @@
...
+    Math.min(
+      srcChunks.filter((c) => c.sign === -1).length,
+      tgtChunks.filter((c) => c.sign === -1).length,
+    ) +
+    Math.min(
+      srcChunks.filter((c) => c.sign === 1).length,
+      tgtChunks.filter((c) => c.sign === 1).length,
+    );
+
+  if (sameSignPairs === 0) {
```
maloyan: **Medium — fallback path doesn't reverse winding; also untested.**

When `sameSignPairs === 0` (e.g. all-CW source vs all-CCW target — the mirrored-glyph case), the function calls `alignChunkListsByLength` once with all chunks pooled regardless of orientation. `alignSubpathPairPoints` only resamples and rotates anchors — it doesn't reverse winding. So opposite-orientation chunks paired here can twist during interpolation rather than morph cleanly.

Also: this branch is never executed by any test in `TransformPairing.test.ts`. All same-sign tests go through the `else` branch.

Suggested fix: detect the all-opposite case and reverse target chunks before length pairing, then add an explicit regression test.

src/core/VMobjectTransformAlignment.ts
```diff
@@ -0,0 +1,420 @@
...
+  }
+}
+
+interface SignedSubpathChunk {
+  chunk: number[][];
+  len: number;
+  sign: 1 | -1;
+}
+
+function signedArea2D(pts: number[][]): number {
```
maloyan: **Low — duplicate orientation oracle.**

This `signedArea2D` is a near-identical copy of `VMobject._signedArea2D` (`VMobject.ts:445`). Two implementations means two places to fix any future bug (self-intersecting polygons, stride change, epsilon handling) and nothing pins their equivalence. Consider extracting to a shared util, or import from `VMobject` directly.

src/core/VMobject.ts
```diff
@@ -316,6 +316,39 @@ export class VMobject extends VMobjectRendering {
...
+    const effective = lengths ?? this.getEffectiveSubpathLengths();
+    if (!effective || effective.length === 0) return undefined;
+
+    const signs: Array<1 | -1> = [];
+    let offset = 0;
+    for (const len of effective) {
+      if (len <= 0 || offset + len > this._points3D.length) return undefined;
+      const chunk = this._points3D.slice(offset, offset + len);
+      const area = VMobject._signedArea2D(chunk);
+      signs.push(area < 0 ? -1 : 1);
```
maloyan, line 337: **Medium — same zero-area-to-CCW issue as `VMobjectTransformAlignment.ts:301`.**

`_signedArea2D` returns `0` for `< 3` anchors / collinear chunks; `area < 0 ? -1 : 1` then collapses every degenerate to CCW. Note this method and the file-private `signedArea2D` in the alignment module are now two parallel oracles — they must agree on the degenerate case or callers will see one bucket assignment from `getSubpathOrientationSigns` and a different one from the alignment module's `inferSign` for the very same chunk.

src/core/VMobject.ts
```diff
@@ -316,6 +316,39 @@ export class VMobject extends VMobjectRendering {
...
+      offset += len;
+    }
+
+    if (offset !== this._points3D.length) return undefined;
```
maloyan: **High — silently returns `undefined` on inconsistent metadata.**

Three separate failure modes — `lengths` empty/missing, any `len <= 0`, `lengths` overrunning `_points3D`, sum mismatch — all collapse into a single `undefined` return with no log and no throw. The caller (`alignVmobjectPair` at `TransformPairing.ts:55-56`) then proceeds with `srcSigns = undefined`, and `alignCompoundPathsForTransform` falls through to `inferSign` per chunk. The result is a *different* sign assignment than the metadata path would have produced, with no diagnostic that metadata was malformed.

Fix: throw on inconsistency. If callers can't supply consistent metadata, that's a bug worth surfacing — and `validateSubpathLengths` already throws elsewhere for the same invariant.

src/core/VMobject.ts
```diff
@@ -316,6 +316,39 @@ export class VMobject extends VMobjectRendering {
...
+
+  /**
+   * Get orientation sign for a specific subpath index: +1 (CCW) or -1 (CW).
+   */
+  getSubpathOrientationSign(index: number): 1 | -1 {
+    const signs = this.getSubpathOrientationSigns();
+    if (signs && index >= 0 && index < signs.length) {
+      return signs[index];
+    }
+    return VMobject._signedArea2D(this._points3D) < 0 ? -1 : 1;
```
maloyan: **Low — wrong fallback masks invalid input.**

For an out-of-range `index`, this returns the orientation of the *whole* point array, treating a compound glyph as one polygon. For a `0` glyph (CCW outer + CW hole) the combined signed area depends on which area dominates — the answer is meaningless. Method also appears unused inside the repo (only the array variant is called).

Fix: throw `RangeError`, return `undefined` (relaxing the return type), or remove the method.

src/animation/transform.test.ts
```diff
@@ -96,14 +109,44 @@ describe('Transform', () => {
...
+          const targetStart = aligned.targetPoints[0];
+          distance = Math.hypot(
+            sourceStart[0] - targetStart[0],
+            sourceStart[1] - targetStart[1],
+            sourceStart[2] - targetStart[2],
+          );
+          break;
+        } catch {
+          // Try next leaf pair.
+        }
```
maloyan: **Medium — empty catch swallows the invariants this regression test was added to enforce.**

This is the headline regression test for the PR's stated goal (MathTex 1→0 first-frame closeness). The `try { … } catch { /* try next leaf pair */ }` swallows every exception from `alignVmobjectPair`, including `validateSubpathLengths` mismatch, `assertSubpathLengthsMatchPoints` throw, and the explicit `alignPoints fallback produced mismatched point counts` error. If a future regression breaks alignment on leaf 0, the test silently skips to leaf 1 and asserts against an unrelated leaf.

Fix: remove the try/catch and let the test fail loudly. If a specific leaf pair is expected to fail, assert that exact case explicitly. As written, the assertion at line 148 protects very little of what the PR description claims.

rambip:
@copilot 

 read the review and fix.
I confirm that modifications to the triangulation algorithm were an error on my side.
For the CW / CCW logic, if the change is too heavy you can give up.

rambip:
@maloyan ready

TODO.md
maloyan, line 1: Delete this file

maloyan (commented):
After removing TODO.md it's good for merge

rambip:
oops sorry !

PR #329, opened 2026-05-09
fix(vmobject): make isNearlyLinear scale-invariant (#310)

### LINKED ISSUE #310: `isNearlyLinear` is wrong for MathTex and create artefacts
## Description
`isNearlyLinear` is wrong for MathTex and create artefacts

## Steps to Reproduce
1. create `MathTex` with text "0"
2. create `MathTex` with text "0" and bigger scale
3. Transform between the 2
4. see the visual effect

## Additional Context
I thought it was solved by my earlier commit (see here: https://github.com/maloyan/manim-web/blob/c3c2cd9451b45006f0bc0f3f59d7185741c00094/src/core/VMobjectGeometry.ts#L177) bu I have seen it recently.

And I think `isNearlyLinear` is still broken

maloyan:
## Summary

Fixes #310. `isNearlyLinear` (src/core/VMobjectGeometry.ts) used a fixed absolute threshold (`0.01` world units) to decide if a cubic Bezier segment is straight enough to skip per-sample evaluation. The same logical glyph at different scales — e.g. a small `MathTex("0")` and a big `MathTex("0")` — therefore classified differently. During a `Transform` between them, sample counts flipped mid-animation and the polyline morphed into rectangular artefacts.

Switching the criterion to `max(perpDist) < TOL_REL * chord` (1% of chord length) makes classification scale-invariant: every uniform scaling of the same control polygon gets the same answer. Existing tests still pass since they use unit-length chords.

## Changes

- `src/core/VMobjectGeometry.ts` — `isNearlyLinear` now compares perpendicular handle distance to chord length (relative tolerance). Degenerate-chord branch additionally checks that handles collapse to the endpoint, so curl/loop Beziers (chord ≈ 0 with bulging handles) are still sampled.
- `src/core/vmobject-coverage.test.ts` — three regression tests pinning scale invariance, the small-curve bug, and the small-linear case.
- `examples/mathtex_zero_scale_morph.{html,ts}` — minimal visual repro of the issue (small "0" → big "0" Transform). Glyph stays curved at every scale post-fix.

## Test plan

- [x] `npx vitest run src/core/vmobject-coverage.test.ts` — 80 pass (3 regression cases now pass; failed against the pre-fix code)
- [x] `npx vitest run` (full suite) — 5873 / 5873 pass
- [x] `npm run lint` — clean
- [x] `npm run typecheck` — clean
- [x] Manual: load `examples/mathtex_zero_scale_morph.html`, click Play — the "0" smoothly grows from `fontSize=24` to `fontSize=240` with the glyph staying curved throughout (no rectangular frames).

### DIFF
--- a/examples/mathtex_zero_scale_morph.html
+++ b/examples/mathtex_zero_scale_morph.html
@@ -1,0 +1,92 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <script>
+      if (new URLSearchParams(location.search).has('embed'))
+        document.documentElement.classList.add('embed');
+    </script>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ManimWeb - MathTex zero scale morph (issue #310 regression)</title>
+    <style>
+      * {
+        margin: 0;
+        padding: 0;
+        box-sizing: border-box;
+      }
+      body {
+        background: #1a1a2e;
+        display: flex;
+        flex-direction: column;
+        justify-content: center;
+        align-items: center;
+        min-height: 100vh;
+        font-family: system-ui, sans-serif;
+      }
+      #container {
+        border: 2px solid #333;
+        border-radius: 8px;
+        overflow: hidden;
+      }
+      .controls {
+        margin-top: 20px;
+        display: flex;
+        gap: 10px;
+      }
+      button {
+        padding: 10px 20px;
+        background: #e94560;
+        border: none;
+        border-radius: 4px;
+        color: white;
+        cursor: pointer;
+        font-size: 14px;
+      }
+      button:hover {
+        background: #ff6b6b;
+      }
+      button:disabled {
+        background: #666;
+        cursor: not-allowed;
+      }
+      html.embed,
+      html.embed body {
+        margin: 0 !important;
+        padding: 0 !important;
+        overflow: hidden;
+        background: #000 !important;
+        width: 100%;
+        height: 100%;
+      }
+      html.embed body {
+        display: flex;
+        justify-content: center;
+        align-items: center;
+      }
+      html.embed .controls,
+      html.embed .buttons,
+      html.embed h1,
+      html.embed #status {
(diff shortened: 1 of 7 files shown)


maloyan:
Pushed a follow-up commit that addresses the chimeric mid-Transform frames between MathTex glyphs (e.g. \`4 → 5\`) reported during review. Three coordinated changes in \`src/core/VMobjectTransformAlignment.ts\`:

1. **Arc-length subdivision** in \`remapBezierCurveCount\`. Previously each cubic was split at equal-\`t\` values, so the i-th anchor of the resampled chain didn't sit at i/N of the perimeter. Now extra splits are allocated to cubics in proportion to their arc length, and each cubic is subdivided at arc-length-uniform \`t\`s. Anchor index now corresponds to a meaningful path-fraction.
2. **Global-SSD rotation with reversal** in \`rotateClosedTargetForBestAlignment\` (replacing \`pickClosestAnchorShift\`). Now that anchor i is at \`i/N\` of perimeter, vertex correspondence between source and target is well-defined, so we exhaustively search every stride-aligned cyclic rotation AND both orientations and pick the rotation with minimum total per-anchor SSD.
3. **Better collapse for unpaired subpaths**. When source has an inner-triangle hole that target doesn't (4 → 5), the missing target subpath used to collapse to its own side's centroid — a meaningless mid-shape point. Now it collapses to the closest anchor on the OTHER side's paired-outer subpath, so a vanishing hole merges into the outer boundary at the most plausible feature.

Also rewires \`VMobject._interpolatePointList3D\` to delegate to the same bezier-aware resampler, so simple non-compound transitions (like 1 → 2) get the same treatment instead of the previous piecewise-linear lerp that silently broke cubic handles.

The existing \`1 → 0\` regression test was pinning the single-anchor heuristic; updated to assert the new invariant (chosen rotation has SSD no larger than any other stride-aligned cyclic rotation). Full suite still 5873 / 5873 pass.

Visually verified by stepping the morph in browser: every mid-frame is now a coherent shape — no detached curls, no rectangular flat tops, no floating-phantom hole.

PR #345, opened 2026-05-12
fix: interpolate scaleVector in VGroup Transform animations

rambip:
## Summary

`Transform` refactor in #319 introduced a bug: `scale` was not interpolated.
This PR fixes it and adds a non-regression test

Brief description of changes.

## Changes
- Add startScale/targetScale to VGroupLeafState interface
- Capture scaleVector in _build() for each child
- Interpolate scaleVector in interpolate() via lerpVectors
- Apply final scaleVector in finish()

- Add test verifying animate.scale ends with correct scaleVector

## Testing
- [x] Tests pass (`npm test`)

# Note

I noticed in #347 that in some case when we scale an object we only update its geometry not its scale. maybe it's related.

### DIFF
--- a/src/animation/transform/PointMorphStrategy.ts
+++ b/src/animation/transform/PointMorphStrategy.ts
@@ -26,6 +26,8 @@ interface VGroupLeafState {
   targetPoints: number[][];
   startPosition: THREE.Vector3;
   targetPosition: THREE.Vector3;
+  startScale: THREE.Vector3;
+  targetScale: THREE.Vector3;
   finalTargetPoints: number[][];
   finalTargetSubpathLengths?: number[];
   startStyle: ChildStyle;
@@ -151,12 +153,24 @@ export class PointMorphStrategy implements MorphStrategy {
     } else child.setTransformSubpathLengths(undefined);
     const startStyle = sourceIsPlaceholder ? captureStyleFaded(tc) : captureStyle(child);
     const targetStyle = targetIsPlaceholder ? captureStyleFaded(sc) : captureStyle(tc);
+    const startScale = new THREE.Vector3(
+      src.leaf.scaleVector.x,
+      src.leaf.scaleVector.y,
+      src.leaf.scaleVector.z,
+    );
+    const targetScale = new THREE.Vector3(
+      tgt.leaf.scaleVector.x,
+      tgt.leaf.scaleVector.y,
+      tgt.leaf.scaleVector.z,
+    );
     return {
       child,
       startPoints,
       targetPoints,
       startPosition: worldToParentLocalPosition(src.worldPosition, src.parentWorldMatrix),
       targetPosition: worldToParentLocalPosition(tgt.worldPosition, src.parentWorldMatrix),
+      startScale,
+      targetScale,
       finalTargetPoints,
       finalTargetSubpathLengths,
       startStyle,
@@ -181,6 +195,7 @@ export class PointMorphStrategy implements MorphStrategy {
           interpolated.push(lerpPoint(leaf.startPoints[i], leaf.targetPoints[i], alpha));
         leaf.child.setPoints(interpolated);
         leaf.child.position.lerpVectors(leaf.startPosition, leaf.targetPosition, alpha);
+        leaf.child.scaleVector.lerpVectors(leaf.startScale, leaf.targetScale, alpha);
         const ss = leaf.startStyle,
           ts = leaf.targetStyle;
         leaf.child.opacity = ss.opacity + (ts.opacity - ss.opacity) * alpha;
@@ -238,6 +253,7 @@ export class PointMorphStrategy implements MorphStrategy {
       for (const leaf of this._vgroupLeafStates) {
         leaf.child.setPoints(leaf.finalTargetPoints);
         leaf.child.position.copy(leaf.targetPosition);
+        leaf.child.scaleVector.copy(leaf.targetScale);
         const ts = leaf.targetStyle;
         leaf.child.opacity = ts.opacity;
         leaf.child.fillOpacity = ts.fillOpacity;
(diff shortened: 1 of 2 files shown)


rambip:
@maloyan Ready for the review.

In particular, I might be missing some other code path where scale interpolation is needed.

rambip:
Closing in favor of #349

src/animation/transform/PointMorphStrategy.ts
```diff
@@ -151,12 +153,24 @@ export class PointMorphStrategy implements MorphStrategy {
...
+    const startScale = new THREE.Vector3(
+      src.leaf.scaleVector.x,
+      src.leaf.scaleVector.y,
+      src.leaf.scaleVector.z,
+    );
+    const targetScale = new THREE.Vector3(
+      tgt.leaf.scaleVector.x,
+      tgt.leaf.scaleVector.y,
+      tgt.leaf.scaleVector.z,
+    );
```
maloyan, line 165: **[P2] Placeholder leaves read `scaleVector=(1,1,1)` and now visibly snap-scale.**

For unmatched VGroup leaves, `pairLeafSnapshotsByIndex` builds the missing side via `makePlaceholderSnapshot()` in `TransformPairing.ts:100`, which returns `{ leaf: new VMobject(), ..., worldScale: reference.worldScale }`. The reference scale is stored in `worldScale`, **not** on the fresh `leaf.scaleVector`, which stays at the default `(1,1,1)`.

This means with the new snapshot code:
- `sourceIsPlaceholder` (fade-in of new leaf): `startScale = (1,1,1)`, `targetScale = tc.scaleVector` ⇒ leaf fades in while scaling from unit to its real size.
- `targetIsPlaceholder` (fade-out of old leaf): `startScale = sc.scaleVector`, `targetScale = (1,1,1)` ⇒ leaf fades out while scaling toward unit.

Both are regressions whenever the leaf is non-unit scaled (e.g. `Transform` where one side adds/removes a scaled child). Before this PR `scaleVector` was simply not touched, so the existing/copied child stayed at its natural size during the cross-fade.

Fix options:
- Skip scale interpolation entirely when `sourceIsPlaceholder || targetIsPlaceholder` (set both to the non-placeholder side, so `lerpVectors` is a no-op).
- Or thread the real scale through `makePlaceholderSnapshot` onto the placeholder leaf (e.g. `leaf.scaleVector.copy(reference.worldScale)` — though world vs. local matters here).

The simplest, lowest-risk fix is the first.

maloyan, line 165: **Nit:** `new THREE.Vector3(v.x, v.y, v.z)` can be replaced with `v.clone()` to match the existing style on lines 117/130. Same below for `targetScale`.

```ts
const startScale = src.leaf.scaleVector.clone();
const targetScale = tgt.leaf.scaleVector.clone();
```

maloyan (commented):
Automated review via `codex review`.

One correctness regression around placeholder leaves (P2) plus a couple of smaller nits on the new code. Otherwise the fix for the matched-leaf case looks right.

src/animation/transform/PointMorphStrategy.ts
```diff
@@ -181,6 +195,7 @@ export class PointMorphStrategy implements MorphStrategy {
           interpolated.push(lerpPoint(leaf.startPoints[i], leaf.targetPoints[i], alpha));
         leaf.child.setPoints(interpolated);
         leaf.child.position.lerpVectors(leaf.startPosition, leaf.targetPosition, alpha);
+        leaf.child.scaleVector.lerpVectors(leaf.startScale, leaf.targetScale, alpha);
```
maloyan, line 198: Once the placeholder-scale issue above is addressed, this `lerpVectors` call should already do the right thing for both matched and unmatched leaves. Worth adding a comment that for placeholder pairs the start and target scales must be equal so this is a no-op (otherwise future readers may re-introduce the regression).

src/core/core-extra.test.ts
```diff
@@ -1820,6 +1825,22 @@ describe('VGroup - extended coverage', () => {
...
+    const a = new VMobject();
+    a.scaleVector.set(1, 1, 1);
+    const vg = new VGroup(a);
+    scene.add(vg);
+
+    await scene.play(vg.animate.scale(2));
+
+    // After animate.scale, child should have correct scaleVector
+    expect(a.scaleVector.x).toBe(2);
+    expect(a.scaleVector.y).toBe(2);
```
maloyan, line 1840: The test only covers a child that started at `(1,1,1)` scaling to `(2,2,2)`. It will not catch the placeholder regression flagged above, since both pairs are matched and start scale is already unit.

Suggest adding (or amending) a case where:
1. The source VGroup has a child with non-unit scale and the target adds a new child (so the new child goes through the `sourceIsPlaceholder` branch).
2. After `scene.play(Transform(source, target))`, the surviving children retain their expected `scaleVector` and the newly-added child ends at its target scale rather than `(1,1,1)`.

That would lock down the regression rather than only the matched-leaf path #319 broke.

src/core/core-extra.test.ts
```diff
@@ -1,5 +1,10 @@
 import { describe, it, expect, vi } from 'vitest';
-import { Mobject, UP, LEFT, RIGHT, UL, UR, DL } from './Mobject';
+import { Mobject, UP, LEFT, RIGHT, UL, UR, DL, registerAnimateProxy } from './Mobject';
+import { Scene } from './Scene';
+import { AnimateProxy } from './AnimateProxy';
+
+// Register animate proxy for tests using animate API
+registerAnimateProxy((mobject) => new AnimateProxy(mobject));
```
maloyan, line 7: **Style:** mixing imports with executable code between them is a bit unusual — the `registerAnimateProxy(...)` call sits between two import blocks. Either move it after all imports, or hoist `registerAnimateProxy` setup into a shared test helper / `beforeAll` if more tests will need the animate API. Not blocking.

Issue #348, opened 2026-05-13
`getCenter` for `VGroup` should use bounding box, not average of centers


rambip:
## Description
`getCenter` returns the average of the centers of the children. This means that for objects such as arrow, the center is biased towards the tip.

## Steps to Reproduce
```
// local geometry: line from x=0 to x=1                                                 
const line = new Line([0, 0, 0], [1, 0, 0]);                                            
                                                                                                                                                          
// a point at x=1                                                                       
const p = new Dot();
p.moveTo([0.9,0,0])                   
                                         
const g = new VGroup(line, p);                                                          
                                                                                       
// then:                                                                                
console.log(g.getCenter());
```

## Expected Behavior
Center should be `0.5`

## Actual Behavior
Center is `0.7`

## Additional Context
It's a blocker for my arrow refactor.

rambip:
Closed by #347

Issue #347, opened 2026-05-13
[Refactor] improve `scale` logic with `normalizeTransform`


rambip:
Currently the logic for scale is pretty fragile. It either scales the scale vector when no point is provided, either the points themselves when a point is provided.

https://github.com/maloyan/manim-web/blob/c2f2128504929fa30e5c0982e0f8db0acbd696c8/src/core/Mobject.ts#L283-L308

It is pretty hard to reason about and it could create bugs in the future.
A better solution would be to scale by default using only the scale vector and updating the position. Then add a `normalizeTransform` method and m object that would reset the transform including position and scale and instead modifying the underlying point data. That way each object should be responsible for the reset of the transform.

maloyan:
## Reproduction

Confirmed the dual-path fragility with a minimal repro on `VMobject`:

```ts
function unitSquare() {
  const v = new VMobject();
  v.setPoints([[-1,-1,0],[-1,0,0],[-1,1,0],[0,1,0],[1,1,0],[1,0,0],[1,-1,0],[0,-1,0],[-1,-1,0]]);
  return v;
}

const a = unitSquare(); a.scale(2);                       // path 1: scaleVector *= 2, points untouched
const b = unitSquare(); b.scale(2, { aboutPoint: [0,0,0] }); // path 2: points *= 2, scaleVector unchanged

a.getBoundingBox().width === b.getBoundingBox().width // ✓ same visible result
a.getPoints()[0]         !== b.getPoints()[0]         // ✗ different internal state
a.scaleVector            !== b.scaleVector            // ✗ different internal state
```

The visible result agrees because Three.js applies `scaleVector` at render. The **internal** state diverges, which surfaces as real bugs the moment any downstream code touches raw points:

```ts
const a = unitSquare();
a.scale(3);                                            // scaleVector = (3,3,3), points still unit
applyFunctionImpl(a, p => [p[0] + 10, p[1], p[2]]);    // shift +10 in point space
a.getCenter();                                         // → [30, 0, 0]  (Three.js multiplies the +10 by scaleVector=3)
//                                                       expected [10, 0, 0]
```

So *which* representation `scale()` writes to silently changes the meaning of every subsequent point-level op (`applyFunction`, `applyMatrix`, custom geometry edits, animations that lerp `getPoints()` directly, etc.). `flip()` has the exact same dual-path shape.

## Proposed plan

Implements the design you sketched: scale becomes pure-transform; point-baking becomes explicit via `normalizeTransform`.

[...]

PR #349, opened 2026-05-13
Refactor scaling and getCenter pipeline, in particular for groups

### LINKED ISSUE #343: `VGroup` rotate logic is wrong
## Description
`rotate` on `VGroup` rotates parent AND children. according to the invariants used in the codebase, it should rotate all childs and not the parent. we should even assert that the transforms of the VGroup are always the identity.

## Steps to Reproduce
1. Create an arrow
2. shift it
3. rotate it

You will see the rotation is not done at the center of the arrow. but instead have weird effects (I will create a MRE soon)

## Additional Context
Any extra info, screenshots, or code snippets.

rambip:
## Summary

This PR adds a mechanism, a method called `normalizeTransform` that is called after animations to reset to a canonical state. This typically is used to transform a `scaleVector` to an updated geometry.

From this new design, I updated a few things, either buggy or hard to reason about:
- `getBounds` now throws an error when there is no point, instead of giving a confusing fallback
- `getCenter` for VGroup uses the center of the bounding box instead of the average of the children centers, like Manim does
- `VGroup` tries to go back to a «position=ORIGIN, scale=1» as a normalization
- `VMObject` does the "morph points directly" in `normalizeTransform` instead of in `scale`

## Changes
 - Added/solidified normalizeTransform() canonicalization flow and used it in geometry-sensitive paths.                                                                                    
 - Mobject.getBounds() now normalizes first and throws on empty Three.js bounds (empty Three.js bounds) instead of silently falling back.                                                  
 - Updated bounds aggregation for container groups to skip empty container children in Group.getChildrenBounds() / VGroup.getChildrenBounds() (fixes empty Table/container edge cases      
 while keeping strict empty-geometry errors).                                                                                                                                              
 - Switched VGroup center semantics to bounding-box center (Manim-style), not average-of-children-centers.                                                                                 
 - Removed custom VGroup.scale() behavior (and simplified Group.scale() to base behavior) so scaling follows the base transform model rather than child-mutation overrides.                
 - Implemented VMobject.normalizeTransform() to bake scaleVector into point geometry, reset scale to identity, and mark geometry/parents dirty.                                            
 - Added/updated tests to match strict bounds + canonical transform semantics:                                                                                                             
     - strict throw expectations for empty bounds paths,                                                                                                                                   
     - group/vgroup scale-model expectations,                                                                                                                                              
     - VGroup transform-anchor identity invariant during point-morph transform.                                                                                                            
 - Added new image regression coverage: src/mobjects/image/image-bounds.test.ts verifies getBounds() width/height scale correctly after scale(2).                                          
 - Added MathTex (SVG) coverage for scale() + normalizeTransform() point correctness (points scale as expected and leaf scaleVector resets to identity).                                   

## Testing
- [x] Tests pass (`npm test`)

## Related Issues
Closes #343 #348 and #347 

## Remaining questions

- Should we also reinitialize rotate for VGroup the same way as the position? Currently not done
 - For transforms on VGroup, should we enforce identity-anchor invariants only in point-morph strategy, or across all transform strategies for consistency?                                
 - Do we want an explicit documented contract for when normalizeTransform() is called automatically vs expected to be called by APIs/animations?
 - should we add `normalizeTransform` to AGENTS.md so that agents have the architecture in mind ?


rambip:
@maloyan ready for review

maloyan (commented):
## Codex automated review

The PR moves several transforms from child-space into parent scale/position state, but the bounds and transform code paths are not updated consistently. The biggest regressions are VGroup transforms to scaled targets no longer baking target scale into child geometry, query-time normalization mutating scale state, and reported centers/bounds diverging from rendered Three.js transforms. Empty-bound behavior is now stricter, but several public callers still invoke getBounds without an empty fallback. I ran npm run typecheck successfully; Vitest could not start in the read-only sandbox because Vite tried to write a .vite-temp config file.

_Posted by `codex review` via gh CLI._

src/animation/transform/PointMorphStrategy.ts
```diff
@@ -119,6 +135,7 @@ export class PointMorphStrategy implements MorphStrategy {
   }
   private _beginVGroup(source: VGroup, target: VGroup): void {
     this._isVGroupTransform = true;
+    assertGroupAnchorsAtIdentity(source, 'begin/source');
```
maloyan, line 141: **[BLOCKER]** This only enforces the source VGroup anchor. After this PR, target.scale(2) leaves scale on target.scaleVector, and _build() only uses tc.getPoints()/tgt.worldPosition; it never applies tgt.worldScale to targetPoints/finalTargetPoints. Transform(new VGroup(circle), group.copy().scale(2)) can finish with unchanged child geometry, which is exactly what the old test was checking. Normalize/bake both source and target VGroups before pairing, or apply the snapshot world scale to the target points, then assert both roots are identity.

src/core/Mobject.ts
```diff
@@ -393,17 +393,16 @@ export abstract class Mobject {
     min: { x: number; y: number; z: number };
     max: { x: number; y: number; z: number };
   } {
-    if (this._threeObject) {
-      const box = new THREE.Box3().setFromObject(this._threeObject);
-      return {
-        min: { x: box.min.x, y: box.min.y, z: box.min.z },
-        max: { x: box.max.x, y: box.max.y, z: box.max.z },
-      };
+    this.normalizeTransform();
```
maloyan: **[MAJOR]** Calling normalizeTransform() from a bounds query makes getBounds() destructive. Callers such as GrowFromEdge.begin copy scaleVector before calling getBounds(); for a scaled VMobject, this line bakes the scale into points and resets scaleVector, then finish restores the pre-baked scaleVector and the object ends up double-scaled. Bounds computation should be non-mutating, or callers must normalize before reading/storing transform state.

src/core/Mobject.ts
```diff
@@ -393,17 +393,16 @@ export abstract class Mobject {
...
+    const obj = this.getThreeObject();
+    const box = new THREE.Box3().setFromObject(obj);
+    if (box.isEmpty()) {
```
maloyan, line 413: **[MAJOR]** The strict throw changes the public behavior for empty but positionable mobjects, while callers still assume getBounds() succeeds. SpiralIn.begin, GrowFromEdge.begin, and the changing animation bound updaters call getBounds() directly; the tests mostly swapped empty VMobject fixtures for PointMobject instead of adding caller fallbacks. Either keep a position-based fallback for empty Object3D bounds or make each caller handle empty bounds explicitly with a clear no-op/position fallback.

src/core/VMobject.ts
```diff
@@ -269,6 +269,32 @@ export class VMobject extends VMobjectRendering {
...
+   */
+  override normalizeTransform(): this {
+    const sx = this.scaleVector.x;
+    const sy = this.scaleVector.y;
+    const sz = this.scaleVector.z;
+    if (sx === 1 && sy === 1 && sz === 1) {
+      return this;
+    }
+
+    for (const p of this._points3D) {
```
maloyan: **[MAJOR]** This bakes only this._points3D. If a VMobject has child VMobjects and a non-identity scaleVector, resetting scaleVector drops the transform that Three.js would have applied to descendants. Recursively bake scale into children, including child positions relative to this origin, before resetting; otherwise skip normalization for VMobjects with children.

src/core/VGroup.ts
```diff
@@ -131,30 +133,87 @@ export class VGroup extends VMobject {
...
-   * Get the center of the group (average of all children centers).
-   * @returns Center position as [x, y, z]
+   * Get axis-aligned bounds that enclose all children.
    */
-  override getCenter(): Vector3Tuple {
+  getChildrenBounds(): {
+    min: { x: number; y: number; z: number };
+    max: { x: number; y: number; z: number };
+  } {
+    this.normalizeTransform();
```
maloyan: **[MAJOR]** normalizeTransform() runs before the empty check. For const vg = new VGroup(); vg.position.set(5, 6, 7); vg.getCenter(), this warns, resets position to zero, and returns [0, 0, 0], so a read loses the empty group's placement. Move empty handling before normalization or preserve position when there are no children.

src/core/VGroup.ts
```diff
@@ -131,30 +133,87 @@ export class VGroup extends VMobject {
...
+  override getCenter(): Vector3Tuple {
+    const b = this.getChildrenBounds();
+    return [(b.min.x + b.max.x) / 2, (b.min.y + b.max.y) / 2, (b.min.z + b.max.z) / 2];
+  }
+
+  /**
+   * Normalize group transform by forwarding direct group translation to children.
+   * This keeps VGroup semantics child-driven and avoids double-counting.
+   */
+  override normalizeTransform(): this {
```
maloyan, line 164: **[MAJOR]** This normalizes direct translation only; scaleVector remains on the VGroup. getChildrenBounds()/getCenter() are computed from child.getBounds() and usually do not include the parent VGroup scale, while rendering and Mobject.getBounds() on the group do. After new VGroup(new Dot({ point: [2, 0, 0] })).scale(2), getCenter() can report the unscaled child center even though the rendered group is scaled. Bake scale into child geometry/positions and reset scaleVector, or compute VGroup center/bounds from the group Three object consistently.

src/core/Group.ts
```diff
@@ -172,11 +206,7 @@ export class Group extends Mobject {
    * @returns this for chaining
    */
   override scale(factor: number | Vector3Tuple): this {
-    for (const child of this.children) {
-      child.scale(factor);
-    }
-    this._markDirty();
-    return this;
+    return super.scale(factor);
```
maloyan, line 172: **[MAJOR]** This changes Group.scale() from child scaling to a parent Three.js transform, but getChildrenBounds()/getCenter() still union each child independently and ignore the group's scaleVector. A scaled Group can render at one center/bounds and report another, which breaks moveTo/nextTo/alignTo after scaling. Either keep child-scaling semantics or make getChildrenBounds() include the group transform.

src/core/Group.ts
```diff
@@ -78,28 +78,62 @@ export class Group extends Mobject {
...
     for (const child of this.children) {
-      const center = child.getCenter();
-      sumX += center[0];
-      sumY += center[1];
-      sumZ += center[2];
+      const isEmptyContainer = 'getChildrenBounds' in child && child.children.length === 0;
+      if (isEmptyContainer) {
+        continue;
+      }
+      const b = child.getBounds();
```
maloyan: **[MAJOR]** This only skips containers that are directly empty. A Group containing new VGroup(new VGroup()) reaches child.getBounds(), creates an empty THREE.Group, and throws, even though getChildrenBounds() on that child would fall back to its position. Use the child's getChildrenBounds() when available, or catch empty bounds, so nested empty containers are handled the same as direct empty containers.

src/mobjects/point/PointMobject.ts
```diff
@@ -40,21 +40,20 @@ export class PointMobject extends PMobject {
...
 
     super({
-      points: [{ position, color, opacity }],
+      points: [{ position: [0, 0, 0], color, opacity }],
       color,
       opacity,
       pointSize: size,
     });
+
+    this.position.set(...position);
```
maloyan, line 49: **[MAJOR]** PointMobject now stores location in Mobject.position, but it still inherits PMobject.shift(), which calls super.shift() and also shifts each internal point. Any Group/VGroup shift or moveTo renders the point at position + point, effectively double-translating it, while getPosition() returns only position. Override shift() so the internal point stays at [0, 0, 0] when position is the source of truth.

src/animation/transform.test.ts
```diff
@@ -1013,24 +1013,14 @@ describe('Transform on VGroup (#206)', () => {
 
     const t = new Transform(group, target);
     t.begin();
-
-    const startPts = circle.getPoints().map((p) => [...p]);
     t.interpolate(1);
```
maloyan, line 1042: **[MAJOR]** This replacement weakens the regression test: target.scale(2) is still set up, but the test no longer verifies that the source child's points or rendered bounds actually scale to the target. With the current PointMorphStrategy, group anchors can remain identity while child geometry never changes. Keep the anchor assertion, but also assert the circle's points or bounds match the scaled target at alpha=1/finish.

src/core/Mobject.ts
rambip: I don't agree. an empty VMobject should always raise an error instead of failing silently. If we want to support specific cases, I'm ok with it, but with a good reason. Currently no test enforces that.

src/core/Group.ts
rambip: I simplified dramatically: just call `getBounds` from parent, which is 100% based on three.js implementation.

rambip:
@maloyan should be good now. Not 100% sure, a new review would be nice.

src/core/normalizeContainerTransform.ts
```diff
@@ -0,0 +1,77 @@
...
+  const ry = container.rotation.y;
+  const rz = container.rotation.z;
+  if (rx === 0 && ry === 0 && rz === 0) return;
+
+  const rot = new THREE.Euler(rx, ry, rz, container.rotation.order);
+  for (const child of container.children) {
+    child.position.applyEuler(rot);
+    if (rx !== 0) child.rotate(rx, [1, 0, 0]);
+    if (ry !== 0) child.rotate(ry, [0, 1, 0]);
+    if (rz !== 0) child.rotate(rz, [0, 0, 1]);
```
maloyan: **[high — correctness] wrong pivot AND wrong order.**

1. **Pivot is wrong.** `child.rotate(rx, [1,0,0])` (and the `ry`/`rz` siblings) call into `rotateVMobjectPoints` (`MobjectPositioning.ts:45`), which when no `aboutPoint` is passed defaults to *the child's own bbox center*. But the container's rotation needs to pivot around the **container origin**. So an off-origin VMobject child (e.g. `Line({start:[0,0,0], end:[1,0,0]})` placed inside a `VGroup` rotated by 90° around Z) gets baked at the wrong location. Concretely: the line's bbox center is `[0.5, 0, 0]`; rotating about that produces `[0.5, -0.5] → [0.5, 0.5]`, not the expected `[0,0] → [0,1]` the user would observe pre-normalization.

   Fix: call `child.rotate(angle, { axis: [...], aboutPoint: [0, 0, 0] })` — the second-argument options form already exists.

2. **Rotation order ignored.** Even with the correct pivot, this decomposes the parent rotation into three sequential axis-angle rotations in fixed X→Y→Z order, but `THREE.Euler` supports `XYZ`, `YXZ`, `ZYX`, `XZY`, `YZX`, `ZXY`. For any container whose `rotation.order` is not `XYZ`, the resulting child orientation will not match the parent's intended rotation.

   Also asymmetric with line 39 — `child.position.applyEuler(rot)` honors the order; the child rotation here doesn't.

   Fix: build a single quaternion from the Euler (`new THREE.Quaternion().setFromEuler(rot)`), then call `child.rotate` once with the equivalent angle/axis, or extend the rotate API to accept a quaternion / matrix.

Add at minimum: (a) a test with an off-origin child and `vg.rotate(Math.PI/2)` followed by `normalizeTransform`, asserting the child geometry matches what direct child rotation would produce; (b) a test with `container.rotation.order = 'YXZ'`.

src/core/normalizeContainerTransform.ts
```diff
@@ -0,0 +1,77 @@
...
+function applyScale(container: ContainerLike): void {
+  const sx = container.scaleVector.x;
+  const sy = container.scaleVector.y;
+  const sz = container.scaleVector.z;
+  if (sx === 1 && sy === 1 && sz === 1) return;
+
+  for (const child of container.children) {
+    child.position.set(child.position.x * sx, child.position.y * sy, child.position.z * sz);
+    child.scale([sx, sy, sz]);
+  }
```
maloyan: **[medium] non-VMobject leaves never canonicalize.** `child.scale([sx,sy,sz])` (with no `aboutPoint`) just multiplies `child.scaleVector`. The recursion at line 74 then calls `child.normalizeTransform()` — which is a no-op on the base `Mobject`. So for any non-VMobject/non-VGroup leaf (or any subclass that didn't override `normalizeTransform`), the multiplied `scaleVector` stays. Visuals look right, but post-condition "all anchors canonical" doesn't hold.

Either (a) bake scale here when the child cannot fold it, or (b) document explicitly that canonicalization is best-effort and only VMobject/VGroup folds scale into points.

src/core/normalizeContainerTransform.ts
```diff
@@ -0,0 +1,77 @@
...
+  container._markDirty();
+}
+
+function applyRotation(container: ContainerLike): void {
+  const rx = container.rotation.x;
+  const ry = container.rotation.y;
+  const rz = container.rotation.z;
+  if (rx === 0 && ry === 0 && rz === 0) return;
+
+  const rot = new THREE.Euler(rx, ry, rz, container.rotation.order);
```
maloyan: **[nit] allocates a `THREE.Euler` per call.** Reuse a module-scoped scratch `Euler` (and assign via `.set(rx, ry, rz, container.rotation.order)`) — `normalizeContainerTransform` runs at every animation begin and is called per child.

src/core/Mobject.ts
```diff
@@ -393,17 +393,15 @@ export abstract class Mobject {
...
-      const box = new THREE.Box3().setFromObject(this._threeObject);
-      return {
-        min: { x: box.min.x, y: box.min.y, z: box.min.z },
-        max: { x: box.max.x, y: box.max.y, z: box.max.z },
-      };
+    const obj = this.getThreeObject();
+    const box = new THREE.Box3().setFromObject(obj);
+    if (box.isEmpty()) {
+      const ctor = (this.constructor as { name?: string } | undefined)?.name ?? 'UnknownMobject';
+      throw new Error(`Mobject.getBounds(): ${ctor} (${this.id}) produced empty Three.js bounds.`);
```
maloyan, line 415: **[high] behavior break: throwing on empty bounds.** Previously `getBounds()` returned a 1×1×1 box around `getCenter()`; now it throws. Anything in user code that called `getCenter()` / `getBoundingBox()` on a freshly constructed mobject (no `_threeObject` yet) used to get a graceful fallback and now throws.

Downstream impact in this PR alone:
- `getCenterImpl` (`MobjectPositioning.ts:155`) propagates the throw — so `nextTo`, `alignTo`, `moveTo` against a newly created mobject can now throw instead of being a no-op-ish fallback.
- `Group.getCenter` / `VGroup.getCenter` paper over this with `isEmpty()` guards, but `isEmpty()` only knows about subclasses that implement it (see `VGroup.ts` duck-type below) — a leaf VMobject with no points is not "empty" by that probe and will still trigger `getBounds()` → throw.

Suggest either (a) keeping the throw but adding an explicit `getCenterOrPosition()` for layout call sites that don't care, or (b) downgrading the throw to a `console.warn` + position fallback and only throwing when explicitly requested. At minimum, call this out in CHANGELOG/docs as a breaking API change.

src/core/MobjectPositioning.ts
```diff
@@ -152,13 +152,8 @@ function rotateWithThreeJS(
...
 export function getCenterImpl(mob: MobjectLike): Vector3Tuple {
-  const obj = mob.getThreeObject();
-  tempBox3.setFromObject(obj);
-  if (!tempBox3.isEmpty()) {
-    tempBox3.getCenter(tempVec3);
-    return [tempVec3.x, tempVec3.y, tempVec3.z];
-  }
-  return [mob.position.x, mob.position.y, mob.position.z];
+  const b = mob.getBounds();
+  return [(b.min.x + b.max.x) / 2, (b.min.y + b.max.y) / 2, (b.min.z + b.max.z) / 2];
```
maloyan, line 156: **[high] `getCenter` now throws via `getBounds`.** This is the universal centroid query — used by `nextTo`, `alignTo`, `moveTo(target)`, `arrange`, etc. Used to silently fall back to `mob.position` when bounds were empty; now it throws a `Mobject.getBounds(): … produced empty Three.js bounds.` error.

For leaves that legitimately have no geometry yet (newly constructed `VMobject`, animation source/target before first `setPoints`), this is a regression. See related comment on `Mobject.getBounds` — the contract change should either be reverted to a position fallback inside `getCenterImpl` (preserve the public API), or every layout helper needs an `isEmpty()` guard.

src/animation/transform/PointMorphStrategy.ts
```diff
@@ -175,6 +195,8 @@ export class PointMorphStrategy implements MorphStrategy {
   }
   interpolate(_animation: Animation, source: Mobject, _target: Mobject, alpha: number): void {
     if (this._isVGroupTransform) {
+      const group = source as VGroup;
+      assertGroupAnchorsAtIdentity(group, `interpolate@${alpha.toFixed(3)}`);
```
maloyan: **[medium] hot-path string allocation.** `assertGroupAnchorsAtIdentity(group, \`interpolate@${alpha.toFixed(3)}\`)` runs every interpolation frame for every active VGroup transform — string + template + `toFixed` per call, even when the assertion passes (the string is built before the function decides whether to throw).

Options: (a) drop the per-frame check entirely — `begin` and `finish` already cover the bookends, and nothing inside `interpolate` mutates `group.position`/`scaleVector`; (b) gate it behind a dev-only flag (e.g. `if (process.env.NODE_ENV !== 'production')`); (c) inline the cheap check and only build the message on failure.

src/animation/transform/PointMorphStrategy.ts
```diff
@@ -119,6 +135,10 @@ export class PointMorphStrategy implements MorphStrategy {
   }
   private _beginVGroup(source: VGroup, target: VGroup): void {
     this._isVGroupTransform = true;
+    source.normalizeTransform();
+    target.normalizeTransform();
```
maloyan: **[medium] silent mutation of the morph `target`.** `target.normalizeTransform()` bakes the user's `target` VGroup S/R/T anchors into its children. If the caller reuses that target VGroup elsewhere (very plausible — `Transform(a, b)` doesn't say "we'll mutate `b`"), they'll observe state churn. At minimum document this in `Transform`/`PointMorphStrategy`. Better: snapshot anchors here and restore them in `finish`, or `target.copy().normalizeTransform()` if that's cheap enough.

(Same concern applies to `source`, but `source` is conventionally consumed by Transform.)

src/core/VMobject.ts
```diff
@@ -269,6 +269,37 @@ export class VMobject extends VMobjectRendering {
...
+      p[0] *= sx;
+      p[1] *= sy;
+      p[2] *= sz;
+    }
+
+    for (const child of this.children) {
+      child.position.set(child.position.x * sx, child.position.y * sy, child.position.z * sz);
+      child.scale([sx, sy, sz]);
+    }
```
maloyan, line 300: **[medium] inconsistent with `normalizeContainerTransform` — no recursion.** This loop scales children (`child.position *= s`, `child.scale([s,s,s])`), but then *doesn't* recurse into `child.normalizeTransform()`. By contrast `normalizeContainerTransform` (line 74) does recurse.

Result: after `parent.normalizeTransform()` on a VMobject, `parent.scaleVector === [1,1,1]` but every child's `scaleVector` is the multiplied value (e.g. the test at `core-extra.test.ts:917` even asserts `child.scaleVector.x === 2` post-normalize). That's a defensible design choice for the leaf case, but it conflicts with the docstring ("matches VGroup normalization semantics") and with the assert-identity contract in `PointMorphStrategy`.

Either call `normalizeContainerTransform(this, { translateChild: ... })` here so semantics line up, or update the docstring to make the asymmetry explicit (and explain why).

src/core/VMobject.ts
```diff
@@ -269,6 +269,37 @@ export class VMobject extends VMobjectRendering {
...
+
+    this.scaleVector.set(1, 1, 1);
+    this._geometryDirty = true;
+    this._markDirtyUpward();
```
maloyan, line 302: **[nit] redundant dirtying.** `_geometryDirty = true; this._markDirtyUpward();` — `_markDirtyUpward()` already sets `_geometryDirty` (and flushes upward); the explicit assignment is redundant.

src/core/Group.ts
```diff
@@ -172,11 +175,7 @@ export class Group extends Mobject {
    * @returns this for chaining
    */
   override scale(factor: number | Vector3Tuple): this {
-    for (const child of this.children) {
-      child.scale(factor);
-    }
-    this._markDirty();
-    return this;
+    return super.scale(factor);
```
maloyan, line 172: **[high — correctness] scale-then-translate compounds the factor.** Old `Group.scale` mutated each child's `scaleVector` and left the parent at identity, so subsequent `shift()`/`moveTo()` (which translate children directly via the per-child loop in this file and via `Mobject.shift`) translated children in the parent's untransformed frame. New `Group.scale` just delegates to `super.scale`, which leaves `scaleVector` on the parent.

Now `Group.shift([dx, dy, dz])` writes `child.position += [dx, dy, dz]` (Mobject.shift), and at render time Three.js multiplies that by `parent.scaleVector`. So:
```ts
const g = new Group(new Dot({ point: [1, 0, 0] }));
g.scale(2);          // parent.scale = [2,2,2], child still at x=1 → renders at x=2
g.moveTo([5, 0, 0]); // computes delta = 5 - getCenter() = 5 - 2 = 3
                     // child becomes x=4, renders at x=8 — overshoots by 3
```
`VGroup` has the exact same regression now that `VGroup.scale` is gone (its `shift` is even more direct: it just calls `child.shift(delta)` per child).

Fix options: (a) call `normalizeTransform()` before applying the translation in `shift`/`moveTo`; (b) divide `delta` by `this.scaleVector` (and rotate by `this.rotation`) before forwarding to children; (c) revert to the old child-mutating `scale`. Whichever you pick, please add a regression test that asserts `g.scale(s).moveTo(p).getCenter() === p`.

src/core/Group.ts
```diff
@@ -77,29 +78,31 @@ export class Group extends Mobject {
...
-      const center = child.getCenter();
-      sumX += center[0];
-      sumY += center[1];
-      sumZ += center[2];
+    const isEmptyContainer = (mob: Mobject): boolean => {
+      const candidate = mob as unknown as { isEmpty?: () => boolean };
+      return typeof candidate.isEmpty === 'function' && candidate.isEmpty();
+    };
+    const hasNonEmptyChild = this.children.some((child) => !isEmptyContainer(child));
+    if (!hasNonEmptyChild) {
```
maloyan, line 90: **[low] duck-typed `isEmpty` repeated in two files.** Same `isEmpty?: () => boolean` cast pattern is in `VGroup.ts:25-28`. Consider promoting to an explicit interface (e.g. `HasIsEmpty`) and putting it next to `MobjectLike` in `MobjectTypes.ts`, then narrow with `'isEmpty' in mob`. Keeps the contract discoverable.

Also: leaves without `isEmpty()` are treated as non-empty here, so a Group containing a single empty `VMobject` (no points, no `_threeObject`) will fall through to `getBounds()` and throw. Worth either treating "no `_threeObject`" as empty by default, or documenting the semantics.

src/core/VGroup.ts
```diff
@@ -6,12 +6,28 @@
...
+   * True when this VGroup (recursively) has no renderable geometry.
+   */
+  isEmpty(): boolean {
+    if (this.children.length === 0) {
+      return true;
+    }
+    return this.children.every((child) => {
+      const candidate = child as unknown as { isEmpty?: () => boolean };
+      return typeof candidate.isEmpty === 'function' ? candidate.isEmpty() : false;
+    });
```
maloyan, line 31: **[low] `isEmpty()` semantics are subtle.**
1. Same duck-typed probe as `Group.ts` — see comment there.
2. A leaf VMobject without points returns `false` here (no `isEmpty` method), so `vg.isEmpty()` returns `false` for `new VGroup(new VMobject())`. Then `getCenter()` calls `getBounds()` which throws. The test `'getCenter throws when a child has empty Three.js bounds'` (`core-extra.test.ts:1846`) actually locks this in. Surprising for users — empty leaves probably should propagate emptiness.
3. If `VMobject.isEmpty()` is added later (treating `_points3D.length === 0` as empty), this VGroup probe will pick it up automatically. Worth adding now to avoid the throw.

src/core/VGroup.ts
```diff
@@ -130,31 +146,33 @@ export class VGroup extends VMobject {
     return this;
   }
 
-  /**
-   * Get the center of the group (average of all children centers).
-   * @returns Center position as [x, y, z]
-   */
   override getCenter(): Vector3Tuple {
-    if (this.children.length === 0) {
+    if (this.children.length === 0 || this.isEmpty()) {
```
maloyan: **[nit] redundant guard.** `this.children.length === 0 || this.isEmpty()` — `isEmpty()` already returns `true` when `children.length === 0`, so the first clause is dead.

src/core/VGroup.ts
```diff
@@ -130,31 +146,33 @@ export class VGroup extends VMobject {
...
+   * This keeps VGroup semantics child-driven and avoids double-counting.
+   */
+  override normalizeTransform(): this {
+    normalizeContainerTransform(this, {
+      beforeTranslate: () => {
+        if (!this._warnedAboutDirectPosition) {
+          console.warn(
+            '[manim-web] Direct VGroup.position mutation detected; forwarding translation to children. Use moveTo()/shift() on VGroup.',
+          );
+          this._warnedAboutDirectPosition = true;
```
maloyan: **[low] one-shot warning is too quiet.** `_warnedAboutDirectPosition` fires a single `console.warn` per VGroup instance, then is permanently silent. If a user is doing `vg.position.set(...)` repeatedly (the failure mode the warning is for), they'll get one warning early in their session and miss it. Either (a) emit every time (small cost since the path is conditional anyway), or (b) make the warning a thrown error in dev mode — the morph strategy already asserts identity, this is the diagnostic that explains *why* the assertion would otherwise fire.

src/mobjects/point/PMobject.ts
```diff
@@ -227,13 +227,9 @@ export class PMobject extends Mobject {
...
-    super.shift(delta);
-    for (const point of this._points) {
-      point.position[0] += delta[0];
-      point.position[1] += delta[1];
-      point.position[2] += delta[2];
-    }
-    return this;
+    // PMobject point coordinates are local-space geometry; translating should
+    // move the mobject transform, not rewrite local point data.
+    return super.shift(delta);
```
maloyan, line 258: **[medium] backward-incompatible change to `PMobject.shift`.** Old behavior: shift updates both transform and the per-point `position` data — so `getPoints()[i].position` reflects world-space after shift. New behavior: only the transform moves; points stay local-space. This is a defensible refactor, but it silently changes what `getPoints()` returns post-shift. The PR removes the test that locked the old behavior in (`point.test.ts: 'shift translates all children'` for `PGroup`) — worth calling out in release notes / migration guide so users who read `getPoints()[i].position` after a shift don't get surprised by stale local coordinates.

src/mobjects/point/PointMobject.ts
```diff
@@ -90,11 +87,12 @@ export class PointMobject extends PMobject {
    * Create a copy of this PointMobject
    */
   protected override _createCopy(): PointMobject {
-    return new PointMobject({
+    const copy = new PointMobject({
       position: this.getPosition(),
       color: this.color,
       opacity: this._opacity,
       size: this._pointSize,
     });
```
maloyan, line 95: **[nit] needless local.** `const copy = new PointMobject({...}); return copy;` can collapse back to `return new PointMobject({...});` — the named local doesn't add anything (the previous version was also a direct return).

maloyan (commented):
Code review (codex + manual) — focusing on correctness, edge cases, performance.

**Two correctness issues worth gating the PR on:**
1. After `Group.scale(...)` / `VGroup.scale(...)` leaves the factor on the parent's `scaleVector`, the in-place child-mutating `shift()`/`moveTo()` get implicitly multiplied by that scale at render time — a `vg.scale(2).moveTo([5,0,0])` chain ends up at world x=8 instead of 5. (See comment on `Group.ts:177` / `VGroup.ts:185`.)
2. `normalizeContainerTransform.applyRotation` calls `child.rotate(angle, axis)` with no `aboutPoint`, which defaults to the child's own bbox center — but the container's rotation needs to be applied around the container origin. Off-origin children get baked at the wrong location after `_beginVGroup()` normalizes a rotated target.

**Other themes:**
- `getBounds()` now throws on empty Three.js bounds — a real behavior break for any caller of `getCenter()`/`nextTo()`/`alignTo()` against a freshly constructed mobject.
- Asymmetry between `VMobject.normalizeTransform` (no recursion) and `normalizeContainerTransform` (recurses) leaves child anchors in non-canonical state.
- Per-frame string allocation for the assertion inside `interpolate` is unnecessary in the hot path.

Line-by-line below.

src/core/Mobject.ts
rambip: I added it as a breaking change for 4.0
Does that seems right ?

src/core/MobjectPositioning.ts
rambip: fixed

rambip:
@maloyan I'm getting to the bottom of this, I ask  you a final review.

rambip:
@maloyan the last refactor I did was pretty big, I expected a few things to be broken. Merging right away was not the best idea IMO.

My modification of `CHANGELOG.md` was clearly contradicting #385 for example.

Am I supposed to fix the issues you opened, or do you plan to do it ? It would really have been simpler to keep this PR open ...

rambip:
Also, note that I did not add the `BREAKING CHANGE` line, so you will have to remember it for the 0.4.0 release

PR #350, opened 2026-05-13
feat: add runtime argument validation for animations and mobjects

### LINKED ISSUE #346: [Usability] Improve argument validation
Currently there is almost no argument validation in the TypeScript code.

This is a huge problem because if the user do not remember what exact arguments are supposed to go for an animation or an object, it will fail silently. It is sometimes very hard to debug.

For example:
- if `angle` is not given to `Rotate`, we get a `NaN` during rotation and the object disappears
- if the user provides more than the expected number of arguments, or if the `config` argument is not a config, we get the error while the animation has started
- for `MathTex`, providing source as the first argument instead of in the options raises an error about `waitForRender`, which is completely unrelated.

So at least we should have a list of required arguments for each animation and object, and we should check that the `config` object is not a specific class. Either with a helper, either with TS logic, either with a method on the parent class.

Moagezaaa:
## Summary

Enable typia runtime assertions where needed and consolidate a repeated rotation-options type. This fixes runtime failures (e.g. missing angle in rotate) and centralizes a common options shape to reduce duplication.

What changed

Runtime checks
- `src/animation/movement/Rotate.ts`: validate options with `typia.assert<RotateOptions>()` and ensure `angle` is present.
- `src/animation/Animation.ts`: validated `AnimationOptions` with `typia.assert<AnimationOptions>()`.
- `src/mobjects/text/MathTex.ts`: uses `typia.assert` to validate required `latex` and options (unchanged intent).

Types / consolidation
- `src/core/MobjectTypes.ts`: added `AxisOrOptions` type alias (`{ axis?: Vector3Tuple; aboutPoint?: Vector3Tuple; aboutEdge?: Vector3Tuple }`) so the shape is defined once.
- `src/core/Mobject.ts`: re-exports `AxisOrOptions` and uses it for `rotate()` signatures where applicable.
- Updated callers to use the shared type: `src/core/MobjectPositioning.ts`, `src/core/AnimateProxy.ts`, `src/core/Group.ts`, `src/animation/movement/Rotate.ts`.

Build / plugin fix
- `vite.browser.config.ts`: added `UnpluginTypia()` so the browser bundle (`dist/manim-web.browser.js`) runs the typia transform (this prevents runtime `typia.assert` calls from throwing in the distributed browser build).

Tests / small fixes
- `src/export/export.test.ts`: increased timeout for the GIF routing test (was timing out under the default 5s; now 15s). Test now completes reliably.

## Testing

- `npm run typecheck` — passes
- `npm test` — full suite passes (all tests green)
- `npm run build:browser` — browser bundle builds and shows `unplugin-typia` transforms applied

## Related Issues

Closes #346


Moagezaaa:
@rambip you can review it now

rambip:
> @rambip review it now

Could you please format the PR in the correct way ?
Summary, changes, closed issue ?

I know it's quite cumbersome to do, but otherwise I can't understand what you did.

Moagezaaa:
@rambip I updated it  . you can check

rambip:
I started to look at your changes. It looks correct, but it only checks a very tiny part of all the options in the code base.
It would be nice to have this kind of thing everywhere, but it can get very hard and cumbersome quickly.

I just discovered [typia](https://typia.io/), which can generate this kind of checks just from the typescript types. We just would have to do:
```ts
import typia, { tags } from "typia";
...
typia.assert<RotateOptions>(options) // raises an error automatically if something is off
```
Documentation [here](https://typia.io/docs/validators/assert/)

Pretty cool, but that would involve changing the development setup. From what I understand it would only require to: 
1. add `typia` to the dependencies and `ts-patch` to the dev dependencies
2. modify the tsconfig.json this way:
```
{
  "compilerOptions": {
    "strict": true,
    "strictNullChecks": true, 
    "skipLibCheck": true,
    "plugins": [
      { "transform": "typia/lib/transform" }
    ]
  }
}
```
3. run `ts-patch install` locally and in the CI
4. modify the vite config:

```
import UnpluginTypia from "@typia/unplugin/vite";
import { defineConfig } from "vite";
 
export default defineConfig({
  plugins: [UnpluginTypia()],
});
```

And that would be enough. Then we just would have to call `typia.assert` everywhere it's needed.

@Moagezaaa what is your opinion on that ? sorry to say this to you only now, but that's probably not the smallest change that you have been working on ...

Moagezaaa:
@rambip Thanks for the suggestion, I didn’t know about Typia before but it actually looks very interesting.
I agree that the current validation approach only covers a subset of the possible runtime issues, and using generated assertions directly from the TypeScript types would definitely make the validation more consistent across the codebase.

And no worries about bringing it up now — I’m glad you mentioned it before we went too far with the manual approach 🙂

rambip:
Feel free to try it, even if you only check a few cases.  would be interested to know if that works

Moagezaaa:
@rambip check and give me feedback

src/animation/movement/Rotate.ts
```diff
@@ -126,5 +129,10 @@ export function rotate(
   angle: number,
   options?: Omit<RotateOptions, 'angle'>,
 ): Rotate {
+  if (typeof angle !== 'number' || !isFinite(angle)) {
```
rambip:
```suggestion
  if (typeof angle !== 'number') {
```

Probably not the best place to check for nan/infinite. Let's focus on the most important, otherwise LLMs will worsen the code every day ...

src/core/Mobject.ts
```diff
@@ -240,7 +241,15 @@ export abstract class Mobject {
       | Vector3Tuple
       | { axis?: Vector3Tuple; aboutPoint?: Vector3Tuple; aboutEdge?: Vector3Tuple },
   ): this {
+    if (typeof angle !== 'number' || !isFinite(angle)) {
```
rambip:
```suggestion
    if (typeof angle !== 'number') {
```

same

src/core/Mobject.ts
```diff
@@ -240,7 +241,15 @@ export abstract class Mobject {
...
+      throw new TypeError('Mobject.rotate: angle must be a finite number');
+    }
+
     if (axisOrOptions && !Array.isArray(axisOrOptions)) {
+      // Validate options using typia (ensures shape and types).
+      typia.assert<{ axis?: Vector3Tuple; aboutPoint?: Vector3Tuple; aboutEdge?: Vector3Tuple }>(
```
rambip: create a type alias to avoid repetition:
```
type AxisOrOptions = { axis?: Vector3Tuple; aboutPoint?: Vector3Tuple; aboutEdge?: Vector3Tuple }
```

rambip:
Seems pretty good, I was able to make it run locally.
But compile time is getting very long unfortunately ...

One solution could be to use 2 config files, without typia locally and for tests, but with typia for build in CI (we mostly care about runtime checks for js bundle anyway ...)

And I discovered something else: typescript seems to be disabled for tests, which may be why I got so many errors while developing. Could you remove `exclude": ["src/**/*.test.ts"]` from the typescript config and see if typechecking passes for the tests ?

rambip:
Dead exports, I'm not sure what's going on ... The build config probably has a problem.
To be honest I'm pretty new to this project, and @maloyan knows a lot more about the build setup.

A few small notes for the future:
1. > @rambip check and give me feedback

    that feels a bit authoritative, I know it's not the intent but "can you check please" is a lot more pleasant

2. When you can, add new commits instead of force-push, it's easier to see what you changed between commits

Moagezaaa:
> Dead exports, I'm not sure what's going on ... The build config probably has a problem. To be honest I'm pretty new to this project, and @maloyan knows a lot more about the build setup.
> 
> A few small notes for the future:
> 
> 1. > @rambip check and give me feedback
>    
>    
>    that feels a bit authoritative, I know it's not the intent but "can you check please" is a lot more pleasant
> 2. When you can, add new commits instead of force-push, it's easier to see what you changed between commits

Thanks! It’s actually my first time contributing to an open-source project, so I really appreciate the feedback.

rambip:
> Thanks! It’s actually my first time contributing to an open-source project, so I really appreciate the feedback.

Congratulations :partying_face: 

It's the start of a great adventure !

vite.config.ts
```diff
@@ -1,9 +1,11 @@
-import { defineConfig } from 'vite';
 import { resolve } from 'path';
+import UnpluginTypia from '@typia/unplugin/vite';
+import { defineConfig } from 'vite';
 import dts from 'vite-plugin-dts';
 
 export default defineConfig({
   plugins: [
+    UnpluginTypia(),
```
maloyan, line 8: The browser bundle build (`vite.browser.config.ts`, invoked by `npm run build:browser` and as part of `npm run build`) doesn't include `UnpluginTypia()`. Since the `typia.assert<...>()` calls only work after the transform runs, the bundled `dist/manim-web.browser.js` (the entry point documented in the README) will throw at runtime as soon as users construct an animation. Would you mind adding the same plugin to `vite.browser.config.ts`?

Side note: the PR description mentions a split-config approach with an `__ENABLE_TYPIA__` flag and new files like `vite.build.config.ts` / `runtimeValidation.ts`, but those don't appear in the diff — the transform is currently enabled everywhere. Worth updating the description to match the final implementation.

maloyan (commented):
Thanks for the contribution! Runtime validation is a nice addition. Left one note below — let me know what you think.

Moagezaaa:
@rambip @maloyan can you check please

rambip:
I think we're doing something wrong here. What we want is:
- Fast, no-typia, typescript-enabled compilation for dev (in particular for examples: we want fast iteration)
- Slower compilation for production builds, with typia enabled. That's what we want to do in CI
- Only one or 2 tests that run only for production builds, that test that typia raises the right errors

The first attempt was wrong, as indicated by @maloyan . But I think the fix is wrong too, the current setup will run the slow compile for dev builds (if I understand correctly)

As I said I'm not a typescript guy, I don't know what's the good move here.

src/core/Mobject.ts
```diff
@@ -240,7 +242,15 @@ export abstract class Mobject {
       | Vector3Tuple
       | { axis?: Vector3Tuple; aboutPoint?: Vector3Tuple; aboutEdge?: Vector3Tuple },
   ): this {
+    if (typeof angle !== 'number') {
+      throw new TypeError('Mobject.rotate: angle must be a finite number');
```
maloyan: Tiny inconsistency: the check above only guards `typeof angle !== 'number'`, but `NaN` and `Infinity` are both typed `number` in JS, so they'll pass through silently despite the "must be a finite number" message. Since the earlier review preferred not to enforce finiteness here, the smallest fix is to soften the message:

```suggestion
      throw new TypeError('Mobject.rotate: angle must be a number');
```

Same wording shows up in `src/animation/movement/Rotate.ts:128`.

src/core/Mobject.ts
```diff
@@ -240,7 +242,15 @@ export abstract class Mobject {
       | Vector3Tuple
       | { axis?: Vector3Tuple; aboutPoint?: Vector3Tuple; aboutEdge?: Vector3Tuple },
```
maloyan: Nice job pulling the shape out into `AxisOrOptions` over in `MobjectTypes.ts` — but this public signature still has the inline object form, which means the two can drift apart. Could you reuse the alias here so all the rotation APIs share one source of truth?

```suggestion
      | AxisOrOptions,
```

(And then the `typia.assert<...>` a few lines below can become `typia.assert<AxisOrOptions>(axisOrOptions)`.)

src/core/Mobject.ts
```diff
@@ -240,7 +242,15 @@ export abstract class Mobject {
...
   ): this {
+    if (typeof angle !== 'number') {
+      throw new TypeError('Mobject.rotate: angle must be a finite number');
+    }
+
     if (axisOrOptions && !Array.isArray(axisOrOptions)) {
```
maloyan: Optional / follow-up — totally fine to defer to keep this PR scoped:

The direct-tuple branch (`Array.isArray(axisOrOptions)`) skips validation entirely, so something like `rotate(Math.PI, [1, 2] as any)` still slips through to `rotateMobject()` and produces `undefined`/`NaN` math. A `typia.assert<Vector3Tuple>(axisOrOptions)` on the array branch would close the gap.

Happy to file a separate issue for this if you'd rather not expand the PR.

.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
         with:
           node-version: 22
       - run: npm install --no-package-lock
+      - run: npx ts-patch install
```
maloyan: Heads-up (not blocking): `ts-patch install` is now in every job — lint, typecheck, test, knip, madge, format. Only the typecheck job actually needs a patched `tsc`; the Vitest + Vite paths get the transform from `@typia/unplugin` directly, and the lint/format/knip/madge jobs don't touch the TS transform at all.

Dropping it from the non-typecheck jobs would save a chunk of CI minutes. Easy follow-up — a maintainer can do this in a separate PR if it's simpler.

maloyan (commented):
Took another pass with help from a code-review tool (Codex) — a few small items below, none of them blocking. First off: thanks for sticking with this through the typia pivot, that was a significant scope shift to absorb mid-review and the result looks clean.

**Small consistency items (worth fixing in this PR):**
- The error messages say "must be a finite number" but the checks only enforce `typeof === 'number'`, so `NaN`/`Infinity` still slip through silently. Easiest fix is just dropping the word "finite" from the messages — keeps things truthful and matches the earlier review preference of not adding a `Number.isFinite` check here.
- `Mobject.rotate()`'s public signature still has the inline `{ axis?...; aboutPoint?...; aboutEdge?... }` shape even though the whole point of the new `AxisOrOptions` alias is to have one source of truth. Swapping it in here finishes the consolidation.

**Optional follow-ups (fine to defer to a separate issue/PR — don't let these block this one):**
- The direct tuple branch in `Mobject.rotate()` (`Array.isArray(axisOrOptions)`) skips the new validation entirely. Real bypass, but a small follow-up.
- `ts-patch install` is run in every CI job; only the typecheck job actually needs it. Easy CI-minute win.

**On the dev-compile-speed / config-split discussion:** I'd suggest landing this as-is (typia transform always on) and filing a separate issue for the dev/CI split. That's an architectural call about the build system that's fairer for a maintainer to own than for a first-time contributor to figure out — you've already done plenty of work here.

Thanks again for the contribution — clean diff, good tests, real bug fix. 🎉

Moagezaaa:
can you check please @maloyan 
Thanks a lot for the guidance and the detailed feedback throughout the review process — I really appreciate it, especially as a first-time contributor 🙂 @rambip @maloyan

Moagezaaa:
Good points — both seem worth tracking separately.

If you’d like, you can open issues for them and assign them to me, and I’ll be happy to work on them after this PR is merged.
@maloyan

maloyan:
Pushed fixes to address review + CI:

- Drop `rootDir: "."` (broke knip's source resolution for `./react` / `./vue` exports → Dead exports cascade)
- Drop the typia tsconfig plugin + ts-patch entirely — `@typia/unplugin` via vite/vitest handles all builds, and `tsc --noEmit` doesn't need the transform to type-check `typia.assert<T>()`. Removes the contributor friction `ts-patch install` introduced.
- Restore `opentype.js ^2.0.0` and `lint-staged ^17.0.4` (accidental stale-base downgrades)
- Drop redundant `npx ts-patch install` from CI jobs
- `Mobject.rotate()`: also validate the tuple branch via `typia.assert<Vector3Tuple>` (the `Array.isArray` path was bypassing validation)

All CI green, 5870 tests passing. Thanks for the contribution and for sticking with the typia pivot — nice runtime validation win for #346. Merging.

Issue #215, opened 2026-04-10
Slides behavior


Wizorx:
## Description
It would be nice for the player to have an option to use the arrows to play the next animation (or the previous one in reverse) so the scene would behave like a slide presentation instead of just using the arrows to jump between segments. In addition to that, there could be an option to loop an animation indefinitely so that it will play while the "next slide" isn't displayed.

## Use Case
It would be a nice addition for people who want to show a story with many animations in a live setting while also allowing for interactive examples. [Manim-Slides](https://manim-slides.eertmans.be/latest/) already achieves some of this, but with no interactivity.

## Proposed Solution
Adding an ``slidesMode`` option to ``PlayerOptions`` should be enough, where the ``play()`` method is called when pressing an arrow and the ``pause()`` method is called when reaching a new animation. A ``repeat()`` method could be added to ``Scene`` to achieve the "slide" with perpetual animations.

Wizorx:
So, is there a way to loop an animation in a slide (i.e. make some animation of the time line play constinuously while the user doesn't signal to move on to the next animation)? I couldn't find anything about ir

maloyan:
## Reproduction

Currently `manim-web` slides mode (`slidesMode: true`) auto-pauses at every segment boundary. There is no way to mark a slide so it keeps replaying until the user presses → to advance. This matches what `manim-slides` calls `next_slide(loop=True)`.

Repro (using existing `examples/slides_mode.html`): each `await scene.play(new Rotate(...))` plays once and pauses. There is no API to make any of those rotates loop continuously while the user is paused on that slide.

Searched the code — looping exists only as a **player-wide** flag (`PlayerOptions.loop`) that restarts the whole timeline once it finishes. There is no per-segment loop, and `MasterTimeline` segment metadata has no `loop` flag (`src/animation/MasterTimeline.ts:10-21`). The slides auto-pause logic in `src/player/Player.ts:438-456` always pauses at `targetSeg.endTime`.

## Proposed fix

Add per-slide looping in slides mode, on by request.

**1. Data model** (`src/animation/MasterTimeline.ts`)
- Add `loop?: boolean` to `Segment` interface.
- Extend `addSegment(animations, opts?: { loop?: boolean })` to set the flag.
- (Optionally extend `addWaitSegment` later if a "looping wait" is ever wanted — skip for now.)

**2. Recorder API** (`RecordingScene` in `src/player/Player.ts`)
[...]

PR #355, opened 2026-05-15
feat(slides): rebuild slidesMode on Slide groups, add nextSlide + autoNext (#215)

maloyan:
## Summary

Refactors slidesMode to mirror [manim-slides' `next_slide(loop, auto_next)`](https://manim-slides.eertmans.be/latest/reference/api.html). Builds on #354.

### API

```ts
player.sequence(async (scene) => {
  // Slide 1 (normal — current behavior preserved if nextSlide() is never called)
  await scene.play(new Create(square));

  // Slide 2: loops until → pressed
  await scene.nextSlide({ loop: true });
  await scene.play(new Rotate(square, { angle: Math.PI * 2, duration: 2, rateFunc: linear }));

  // Slide 3: two plays grouped into one slide, advances automatically when done
  await scene.nextSlide({ autoNext: true });
  await scene.play(new FadeIn(label));
  await scene.play(new Indicate(label));

  // Slide 4: normal
  await scene.nextSlide();
  await scene.play(new FadeOut(square));
});
```

- **`scene.nextSlide(opts?)`** — boundary marker (was `scene.loopPlay` in #354, removed).
- **`{ loop: true }`** — slide replays until →.
- **`{ autoNext: true }`** — slide advances into the next slide automatically when done. Final autoNext slide just pauses.
- **Multi-play slides** — every `play()/wait()` between two `nextSlide()` calls forms one slide.
- **Back-compat**: if `nextSlide()` is never called, each segment becomes its own slide → existing per-play slidesMode works unchanged (verified: existing `examples/slides_mode.html` still passes through this path).

### Internals

- Introduces `Slide` in `MasterTimeline` grouping N consecutive `Segment`s: `{ index, startSegmentIndex, endSegmentIndex, startTime, endTime, loop, autoNext }`.
- Drops `Segment.loop` (was per-segment in #354, internal-only) — loop is now a `Slide` property.
- `MasterTimeline.beginSlide(opts)` records boundaries; `finalizeSlides()` builds the slide list at end of `sequence()`. Idempotent. Throws on zero-duration looping slides.
- Player slidesMode state renamed: `_slidesTargetSlideIndex`, `_activeLoopSlideIndex`. `nextSegment`/`prevSegment` keep names + new `nextSlide`/`prevSlide` aliases. Navigation operates on slides — multi-play slides are atomic.
- `_startLoop` slide-boundary branch: `loop` (rewind) > `autoNext` (advance into next slide, keep playing) > default (pause). Final `autoNext` falls through to pause.
- PlayerUI scrubber markers now mark slide boundaries (the navigable unit).

### Out of scope (follow-up)

- Slide titles / presenter notes.
- `auto_next` for non-slidesMode (today the flag is slidesMode-only).

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npx vitest run` — **5909 passed**
- [x] Browser verify: slide 0 plays then pauses; slide 1 loops bounded inside `[startTime, endTime]`; pressing → exits loop into slide 2 (multi-play autoNext) which then auto-advances through slide 3 to the final pause.

Plan + review trail: #215, #354.

### DIFF
--- a/examples/slides_mode_loop.html
+++ b/examples/slides_mode_loop.html
@@ -3,7 +3,7 @@
 <head>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <title>ManimWeb - Slides Mode with Looping Slide</title>
+  <title>ManimWeb - Slides Mode (loop / autoNext / multi-play)</title>
   <style>
     * { margin: 0; padding: 0; box-sizing: border-box; }
     body {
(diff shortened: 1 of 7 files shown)


Wizorx:
One more question: is there a way to add an interactive slide? What I mean is a slide that when reached, essentially pauses the presentation, but the objects can still be moved around and interacted with.

PR #394, opened 2026-05-20
fix(#390) fix(#391) normalizeTranform fixes for groups

### LINKED ISSUE #390: VGroup.isEmpty() over-classifies VMobject containers without own points
## Problem

`VGroup.isEmpty()` (`src/core/VGroup.ts:22-31`) falls back to `child.getPoints().length === 0` for `VMobject` children. Renderable container-style VMobjects (e.g. `CurvesAsSubmobjects`, `Code`, `DecimalNumber`) that draw exclusively through submobjects/meshes but have no own control points are classified as empty.

Downstream, `VGroup.getCenter()` short-circuits to the group `position` (skipping `getBounds()`), so placement APIs like `moveTo` / `arrange` / `next_to` are wrong for groups whose only children are these renderable-but-pointless VMobjects.

## Repro sketch

```ts
const code = new Code(...);  // VMobject, renders via submobjects, _points3D = []
const g = new VGroup(code);
g.moveTo([1, 1, 0]);  // ignores code bounds
```

## Suggested fix

Treat a VMobject as empty only when *both* own points and submobjects are empty (recursive `isEmpty` check), or always consult `getBounds()` when in doubt.

Found by codex review of #349.

### LINKED ISSUE #391: Group rotation normalize drops VMobject child.position offset
## Problem

When `Mobject._normalizeContainerRotation()` (`src/core/Mobject.ts:610-620`) bakes a container's rotation into its children, it calls `child.rotate(angle, { aboutPoint: [0,0,0] })`.

For VMobject children, this dispatches to `rotateVMobjectPoints` (`src/core/MobjectPositioning.ts:33-39`), which rotates only `_points3D` and ignores `child.position`. A child with a non-zero `position` (e.g. a `Line` shifted before being grouped, then the group rotated) keeps its old offset after the parent rotation is cleared — visible jump on next `getCenter`/`moveTo`/`shift`.

## Repro sketch

```ts
const line = new Line(...).shift([2, 0, 0]);
const g = new VGroup(line);
g.rotate(Math.PI / 2);   // parent rotation
g.normalizeTransform();  // bakes; line.position still [2,0,0] but expected ~[0,2,0]
```

## Suggested fix

In the normalize path, rotate `child.position` explicitly as well, or take a non-VMobject rotation code path that affects both points and position.

Found by codex review of #349.

rambip:
## Summary                                                                                                                                                                               
                                                                                                                                                                                           
 Fixes two transform/geometry correctness bugs in grouped objects:                                                                                                                         
                                                                                                                                                                                           
 - #391: normalizeTransform() now correctly bakes deferred group rotation into VMobject child position offsets (not just VMobject points).                                                 
 - #390: emptiness checks are now consistent/recursive via base Mobject.isEmpty(), so VGroup no longer misclassifies point-less VMobject containers with renderable descendants as empty.  
                                                                                                                                                                                           
 ## Changes                                                                                                                                                                               
                                                                                                                                                                                           
 - src/core/Mobject.ts                                                                                                                                                                     
     - In container-rotation normalization, rotate VMobject child position offsets while applying Euler steps.                                                                             
     - Add base isEmpty(): boolean (conservative default).                                                                                                                                 
 - src/core/VMobject.ts                                                                                                                                                                    
     - Add override isEmpty():                                                                                                                                                             
           - non-empty if own points exist                                                                                                                                                 
           - otherwise recursively checks children emptiness.                                                                                                                              
 - src/core/VGroup.ts                                                                                                                                                                      
     - Simplify isEmpty() to rely on child.isEmpty() recursively.                                                                                                                          
 - src/core/core-extra.test.ts                                                                                                                                                             
     - Add regression test for #391 (shifted Line in VGroup keeps correct offset after rotate + normalize).                                                                                
     - Add regression test for #390 (point-less VMobject container with renderable descendant is treated as non-empty; getCenter/moveTo use real bounds).      

## Testing
- [x] Tests pass (`npm test`)

## Related Issues
Closes #390 
Closes #391

### DIFF
--- a/src/core/Group.ts
+++ b/src/core/Group.ts
@@ -77,17 +77,12 @@ export class Group extends Mobject {
     return this;
   }
 
+  override isEmpty(): boolean {
+    return this.children.length === 0 || this.children.every((child) => child.isEmpty());
+  }
+
   override getCenter(): Vector3Tuple {
     if (this.children.length === 0) {
-      return [this.position.x, this.position.y, this.position.z];
-    }
-
-    const isEmptyContainer = (mob: Mobject): boolean => {
-      const candidate = mob as unknown as { isEmpty?: () => boolean };
-      return typeof candidate.isEmpty === 'function' && candidate.isEmpty();
-    };
-    const hasNonEmptyChild = this.children.some((child) => !isEmptyContainer(child));
-    if (!hasNonEmptyChild) {
       return [this.position.x, this.position.y, this.position.z];
     }
 
--- a/src/core/Mobject.ts
+++ b/src/core/Mobject.ts
@@ -403,6 +403,14 @@ export abstract class Mobject {
 
   // ── Positioning & Bounding Box ───────────────────────────────────
 
+  /**
+   * Whether this mobject has no renderable geometry.
+   * Base default is conservative: non-empty unless specialized subclasses decide otherwise.
+   */
+  isEmpty(): boolean {
+    return false;
+  }
+
   getCenter(): Vector3Tuple {
     return getCenterImpl(this);
   }
--- a/src/core/Mobject.ts
+++ b/src/core/Mobject.ts
@@ -4,6 +4,7 @@ import {
   type Vector3Tuple,
   type MobjectStyle,
   type AxisOrOptions,
+  isVMobjectLike,
   UP,
   DOWN,
   LEFT,
@@ -34,6 +35,8 @@ import { axisVectorFromEulerKey } from '../utils/axis';
 let animateProxyFactory: ((mobject: Mobject) => any) | null = null;
 
 const SCRATCH_EULER = new THREE.Euler();
+const SCRATCH_CHILD_POS = new THREE.Vector3();
+const SCRATCH_CHILD_QUATERNION = new THREE.Quaternion();
 
 interface NormalizeContainerOptions {
   translateChild?: (child: Mobject, dx: number, dy: number, dz: number) => void;
@@ -611,8 +614,26 @@ export abstract class Mobject {
       const eulerAxis = axis as 'X' | 'Y' | 'Z';
       const angle = eulerAxis === 'X' ? rx : eulerAxis === 'Y' ? ry : rz;
       if (angle !== 0) {
+        const rotationAxis = axisVectorFromEulerKey(eulerAxis);
+
+        // VMobject rotation transforms points directly and does not rotate child.position.
+        // During container normalization we must also bake the positional offset.
+        if (isVMobjectLike(child) && child._points3D.length > 0) {
+          SCRATCH_CHILD_POS.set(child.position.x, child.position.y, child.position.z);
+          SCRATCH_CHILD_QUATERNION.setFromAxisAngle(
+            SCRATCH_CHILD_POS.set(rotationAxis[0], rotationAxis[1], rotationAxis[2]).normalize(),
+            angle,
+          );
+          SCRATCH_CHILD_POS.set(
+            child.position.x,
+            child.position.y,
+            child.position.z,
+          ).applyQuaternion(SCRATCH_CHILD_QUATERNION);
+          child.position.set(SCRATCH_CHILD_POS.x, SCRATCH_CHILD_POS.y, SCRATCH_CHILD_POS.z);
+        }
+
         child.rotate(angle, {
-          axis: axisVectorFromEulerKey(eulerAxis),
+          axis: rotationAxis,
           aboutPoint: [0, 0, 0],
         });
       }
--- a/src/core/VGroup.ts
+++ b/src/core/VGroup.ts
@@ -15,20 +15,8 @@ export class VGroup extends VMobject {
   /**
    * True when this VGroup (recursively) has no renderable geometry.
    */
-  isEmpty(): boolean {
-    if (this.children.length === 0) {
-      return true;
-    }
-    return this.children.every((child) => {
-      const candidate = child as unknown as { isEmpty?: () => boolean };
-      if (typeof candidate.isEmpty === 'function') {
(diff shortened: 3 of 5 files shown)


src/core/Group.ts
```diff
@@ -77,20 +77,15 @@ export class Group extends Mobject {
     return this;
   }
 
+  override isEmpty(): boolean {
+    return this.children.length === 0 || this.children.every((child) => child.isEmpty());
+  }
+
   override getCenter(): Vector3Tuple {
     if (this.children.length === 0) {
```
maloyan, line 85: `getCenter()` only short-circuits on `children.length === 0`, but this PR added a recursive `Group.isEmpty()` just above (line 80). A recursively-empty group like `new Group(new VGroup(new VGroup()))` now falls through to `getBounds()` and throws `empty Three.js bounds`, while `new Group()` returns the position — and `VGroup.getCenter()` (VGroup.ts:141) gracefully guards with `this.isEmpty()`.

Suggest using the new contract for consistency:

```ts
if (this.isEmpty()) {
  return [this.position.x, this.position.y, this.position.z];
}
```

That also makes the changed test no longer need to expect a throw. If throwing *is* the intended behavior, then `VGroup.getCenter()` should be aligned to match — right now the two container types disagree on identical empty trees.

src/core/Mobject.ts
```diff
@@ -611,8 +622,26 @@ export abstract class Mobject {
       const eulerAxis = axis as 'X' | 'Y' | 'Z';
       const angle = eulerAxis === 'X' ? rx : eulerAxis === 'Y' ? ry : rz;
       if (angle !== 0) {
+        const rotationAxis = axisVectorFromEulerKey(eulerAxis);
+
+        // VMobject rotation transforms points directly and does not rotate child.position.
+        // During container normalization we must also bake the positional offset.
+        if (isVMobjectLike(child) && child._points3D.length > 0) {
+          SCRATCH_CHILD_POS.set(child.position.x, child.position.y, child.position.z);
```
maloyan, line 630: This line is dead — `SCRATCH_CHILD_POS` is overwritten on line 632 (`.set(rotationAxis...)`) before it is ever read, and the position is re-set again on line 635. Safe to delete.

Readability nit: consider a dedicated scratch vector for the axis rather than reusing `SCRATCH_CHILD_POS` for both the axis and the position. The reuse is correct (`setFromAxisAngle` reads the axis synchronously, so no aliasing into the quaternion), but it reads as if there were.

src/core/Mobject.ts
```diff
@@ -611,8 +622,26 @@ export abstract class Mobject {
       const eulerAxis = axis as 'X' | 'Y' | 'Z';
       const angle = eulerAxis === 'X' ? rx : eulerAxis === 'Y' ? ry : rz;
       if (angle !== 0) {
+        const rotationAxis = axisVectorFromEulerKey(eulerAxis);
```
maloyan, line 625: Pre-existing concern, now also affecting position offsets: the enclosing loop (line 621) iterates `this.rotation.order` in forward order, applying one quaternion per axis. But a Three.js `Euler` composes to `qX·qY·qZ`, which applied via sequential `applyQuaternion` needs the axes in *reverse* of `order`. For a genuine multi-axis deferred rotation (e.g. `order='ZXY'` with non-zero x/y/z), both `child.rotate(...)` and the new `child.position` baking will diverge from `parent.rotation`, so `normalizeTransform()` can shift world geometry.

Verified with Three.js: per-axis quaternions in reverse of `order` match `Vector3.applyEuler(euler)` to ~1e-15; forward order is off by ~0.3–0.6.

The new offset-baking is *consistent* with the existing `child.rotate` loop, and single-axis rotations (the #391 case) are unaffected — so this is not a regression. But the added "respects non-XYZ Euler order" test validates against this same loop rather than against `applyEuler`, so it won't catch the mismatch. Worth a follow-up issue, or reverse the iteration order here.

src/core/VGroup.ts
```diff
@@ -15,20 +15,8 @@ export class VGroup extends VMobject {
...
-      const candidate = child as unknown as { isEmpty?: () => boolean };
-      if (typeof candidate.isEmpty === 'function') {
-        return candidate.isEmpty();
-      }
-      if (child instanceof VMobject) {
-        return child.getPoints().length === 0;
-      }
-      return false;
-    });
+  override isEmpty(): boolean {
```
maloyan, line 18: Minor: this override is now redundant. `VGroup extends VMobject`, and the inherited `VMobject.isEmpty()` already returns `children.every(c => c.isEmpty())` when there are no own points (`[].every()` is `true`, covering the empty case). Since a VGroup never has its own `_points3D`, the two are functionally identical — you could drop the override and inherit. Not blocking.

maloyan (commented):
## Code review — PR #394

Reviewed the transform/emptiness fixes for #390 and #391 with Claude + Codex (gpt-5.5). The core approach is sound: the `isVMobjectLike(child) && child._points3D.length > 0` guard in `_applyEulerSteps` exactly mirrors the branch in `rotateMobject` that rotates points but leaves `position` untouched — so baking the offset there is the right place. Targeted tests pass locally (`core-extra.test.ts`, 360 tests).

**Discussion**
- `Group.getCenter()` is now inconsistent with the new recursive `isEmpty()` contract and with `VGroup.getCenter()`.
- `_applyEulerSteps` bakes multi-axis rotations in the wrong Euler order (pre-existing; this PR extends it to position offsets).

**Minor**
- Dead line in the new offset-baking block.
- `VGroup.isEmpty()` override is now redundant.

Single-axis rotation (the actual #391 case) is correct — the multi-axis concern is orthogonal and can be a follow-up. Details inline.

rambip:
Closing in favor of #417

Issue #413, opened 2026-05-26
Property-based tests and contracts


rambip:
Like I presented in my previous issue, test coverage is misleading.

It would be a lot more valuable to write property-based tests, for example with [fast-check](https://github.com/dubzzz/fast-check)

Another very valuable addition would be to add corresponding contracts (properties you want about a function) directly in the jsdocs. It could look like:
```ts
/**
 * @pre array.length > 0
 * @pre array.every((x, i) => i === 0 || x >= array[i - 1])
 * @post (result) => result === -1 || array[result] === key
 */
function binarySearch(array: number[], key: number): number {
  let lo = 0, hi = array.length - 1;
  while (lo <= hi) {
    const mid = (lo + hi) >> 1;
    if (array[mid] === key) return mid;
    if (array[mid] < key) lo = mid + 1;
    else hi = mid - 1;
  }
  return -1;
}
```

Then, the agent would translate it to a test and back.
I noticed the agent finds bugs a lot faster when invariants are written this way.

PR #417, opened 2026-05-29
Normalize transform

### LINKED ISSUE #390: VGroup.isEmpty() over-classifies VMobject containers without own points

### LINKED ISSUE #391: Group rotation normalize drops VMobject child.position offset

### LINKED ISSUE #418: `applyFunction` does not work on `PMobject`
## Description
`applyFunction` does not work on `PMobject`

## Steps to Reproduce
1. create `PMobject` with one single points
2. apply `ApplyFunction` anim
3. see no change

## Additional Context
Will be fixed in #417

rambip:
## Summary
This PR is **GIGANTIC**. I feel only slightly sorry for that, this rewrite was needed.

I rewrote the entire core geometric implementation. Basically, I say that `Group`, `VGroup`, `PMobject`, `Circle`, `Arrow` (etc) override specific methods of `MObject` with a different logic and different conventions each time. Sometimes, they were not documented. Sometimes, there were spagetti-code fixes in the transform themselves. And along the way I found some subtle bugs being there for ages.

## Changes
-  too much for a review. You will need to trust me. I will describe all changes in the next messages

## Testing
- [x] Tests pass (`npm test`)

## Related Issues
Closes #390
Closes #391 
Closes #418

## Notes

I will shortly do a follow-up PR to remove a lot of tests and instead add #413 
If there are any bugs, I'm very confident I will find them at this point.


rambip:
# Idea behind `normalizeTransform`

Since we use `threejs` scene graph logic to represent object, a single scene can be represented by multiple different (but equivalent) groups. For example let's say you want to represent "a group with 2 dots, one at (1,1) and the other at (1,-1). You could represent it in these 2 ways:

image

Code for the first version:
```
const group = new VGroup();
group.add(new Dot({ point: [1,  1, 0] }));
group.add(new Dot({ point: [1, -1, 0] }));
```

Code for the second version:
```
const group = new VGroup();
group.position.set(1, 0, 0);
group.add(new Dot({ point: [0,  1, 0] }));
group.add(new Dot({ point: [0, -1, 0] }));
```

The original code used the first representation: always set position to origin for groups.
But then, the implementations of `scale` and `rotate` where hard because you had to first find the center, and derive  everything from it. That was a lot of duplicated code because subclasses reimplemented this logic.

Plus, if you shifted the position of the group, all point coordinates were off.

I realized that it was a lot easier to use the second convention: always pick the center of the points as the group position. this way you an just set `scale` and `rotation`, it's basically free !
The only problem is that this contract (position == getLocalCenter) can be out of sync. This is the main goal of `normalizeTransform`: go back to a clean transform state, where children are centered around position.

This is even more general in fact: `normalizeTransform` will:
- translate children so that they are centered
- reset scale and rotation to zero
- call `normalizeTransform` on the children
- update `_points3d` or `points` depending on the class


This makes the implementation of most shapes simpler. Let's say you want to create a blue arrow from origin to (2, 1):
image

You can start by creating it in local space:
image

Then moving it using the rotation and scale properties:
image


And at the end calling `normalizeTransform`

image


## Call sites

We call `normalizeTransform`:
- when we create a `group`
- before any animation

# `PMobject`

The same convention is used for PMobjects: position is the center of all given points. This means that for `PointMobject`, we don't have any custom logic: it's own local center is it's position, and `point` stays 0. 

# `Circle`

I removed the `scale` override of circle. Instead, I just have to do this:

```ts
  override normalizeTransform(): this {
    const s = this._getUniformScaleFactor(); // check that scale[0]==scale[1]==scale[2]
    this._radius *= s;
    super.normalizeTransform();
    return this;
  }
```

It's less code, and more elegant.

# Points: `World` vs `Local`

One complexity that is added with our convention is that `_point3d` does not correspond to the position in world space. The user might want to get the local points / center, or the global ones. And they should match if no parent.

For this reason, I added a consistent `getLocal` equivalent of all `get` functions:
- `getLocalCenter`
- `getLocalPoints`

For consistency, I had to rename the current `PMobject.getPoints` to `PMobject.getPointsData` to match the style.

It's the only breaking change for point access I think, worth checking that all examples are updated in the correct way.

# Contracts

In order to document the properties of the different functions, I added contracts as mentioned in #413 

I updated AGENTS.md so that they can create new ones with a good style.

That will be very valuable when implementing property-based tests.

# Other fixes
- now `PMobject` center uses the bounding box instead of the average. `centerOfMass` was added.

rambip:
@maloyan what's your view on this ?
Do you trust me enough to merge this change ?

Maybe you could at least check that most examples look ok ?
I tried to keep as much tests as possible but some of them did not make sense with the new implementation.

I'm vey confident that the codebase will be easier to maintain after this point.

src/mobjects/geometry/Dot.ts
```diff
@@ -110,7 +83,7 @@ export class Dot extends Circle {
    */
   protected override _createCopy(): Dot {
     return new Dot({
-      point: this._point,
+      point: this.getPoint(),
       radius: this.getRadius(),
```
maloyan: **[MAJOR] `Dot._createCopy` double-applies the scale.** `getRadius()` now returns `this._radius * _getUniformScaleFactor()` (it folds in the deferred `scaleVector`). But the base `Mobject.copy()` also copies `scaleVector` onto the clone, so seeding the new `Dot` with `radius: this.getRadius()` gives the copy an effective radius of `r·s·s` whenever the dot has a pending (non-normalized) scale. `Circle._createCopy` avoids this by passing the raw `this._radius` — `Dot` should do the same.

src/mobjects/geometry/Circle.ts
```diff
@@ -75,51 +75,55 @@ export class Circle extends VMobject {
...
+    return sx;
+  }
+
   /**
-   * Get the radius of the circle
+   * Get the radius of the circle (includes deferred uniform scale).
    */
   getRadius(): number {
-    return this._radius;
+    return this._radius * this._getUniformScaleFactor();
```
maloyan, line 127: **[MAJOR] `getRadius()`/`setRadius()` disagree on whether scale is included.** `getRadius()` returns `_radius * _getUniformScaleFactor()`, but `setRadius(v)` stores `v` raw and regenerates at the raw radius. So `setRadius(getRadius())` is not idempotent (`r·s` → `r·s·s`) when a deferred `scaleVector` is present, and `getCircumference`/`getArea`/`pointAtAngle` (all via `getRadius`) disagree with what `setRadius` last set. Either divide the scale out in `setRadius`, or have both accessors agree on the convention. (This getter change is also what makes `Dot._createCopy` double-scale.)

src/mobjects/geometry/Circle.ts
```diff
@@ -135,80 +139,60 @@ export class Circle extends VMobject {
...
-   * Shift the circle by updating its center point and regenerating geometry.
-   * Does not change Mobject.position to avoid double-counting in THREE.js hierarchy.
-   */
-  override shift(delta: Vector3Tuple): this {
-    this._centerPoint[0] += delta[0];
-    this._centerPoint[1] += delta[1];
-    this._centerPoint[2] += delta[2];
-    this._generatePoints();
-    this._markDirty();
+    this.position.set(value[0], value[1], value[2]);
```
maloyan, line 157: **[MAJOR] `setCircleCenter` mutates `position` without `_markDirty()`.** Per the `_dirty` invariant (a clean node ⇒ clean subtree, so `_syncToThree`/`_syncWorldMatrices` skip clean nodes), every other position mutator (`moveTo`, `shift`, `_rotateAboutParentLocalPoint`, …) calls `_markDirty()`. This setter changes `this.position` but never marks the node dirty, so the THREE object won't re-sync and the circle renders at its old center until something else dirties it. Add `this._markDirty();` before `return this;`.

maloyan (commented):
## Automated review of #417 (Normalize transform)

Reviewed with a multi-agent pass (per-subsystem readers + adversarial verification) and a second independent pass with `codex`. CI is green and the local↔world split is implemented cleanly in the core (the `Mobject` world = parentChain ∘ local composition, the `MobjectState`/`PointBounds` round-trips, and the `Group`/`VGroup` gutting all look good). Nice cleanup overall.

I'm **not approving** — several real correctness bugs survived verification. Most are *latent*: they don't fire in the suite because no test transforms/parents the object before querying it, but they are regressions vs `main` for transformed/parented objects. Inline comments below on:

- `Dot._createCopy` double-applies scale (interacts with the new `Circle.getRadius`)
- `Circle.getRadius`/`setRadius` scale asymmetry
- `Circle.setCircleCenter` missing `_markDirty()` (stale render)
- `Group.getCenter` can throw on an all-empty group (inconsistent with `VGroup`)
- `Line.getStart/getEnd` world-space vs the sibling accessors still local-space
- `Brace` local key-points mixed with a world-space center

### Also worth verifying (lower confidence / flagged by codex — not blocking)
- **`Mobject.moveTo` (~281):** computes a world-space delta then applies it via `shift()`. Correct for unparented objects; please confirm it's right for an object parented under a scaled/rotated parent.
- **`VGroup.getLocalPoints` (452) + inherited `setPoints`:** the getter aggregates children, but there's no matching setter override, so a read-modify-write (e.g. `Homotopy`/`PhaseFlow` run on a VGroup) writes the combined points onto the container itself instead of the children — a silent no-op.
- **`MoveAlongPath`/`Homotopy` `getPoints()`→`getLocalPoints()`:** if the path / target carries its own transform, it's now sampled/applied in local space — confirm that's intended.
- **`VMobject.getPoints()` semantic flip (local→world):** a silent behavioral break for any external caller relying on the old local semantics.
- **`Mobject.rotate` (139):** `axis.normalize()` mutates a caller-supplied vector in place — clone before normalizing.

Happy to re-review once these are addressed.

src/core/Group.ts
```diff
@@ -77,101 +81,19 @@ export class Group extends Mobject {
...
-    const isEmptyContainer = (mob: Mobject): boolean => {
-      const candidate = mob as unknown as { isEmpty?: () => boolean };
-      return typeof candidate.isEmpty === 'function' && candidate.isEmpty();
-    };
-    const hasNonEmptyChild = this.children.some((child) => !isEmptyContainer(child));
-    if (!hasNonEmptyChild) {
-      return [this.position.x, this.position.y, this.position.z];
-    }
-
     const b = this.getBounds();
```
maloyan, line 103: **[MAJOR] `getCenter()` can now throw for an all-empty group.** This refactor removed the old `hasNonEmptyChild` fallback (which returned `[position]` when no child had geometry). Now, when `children.length > 0` but every child is empty (e.g. async Text/MathTex glyph containers before they populate, or nested empty VGroups), execution reaches `getBounds()`, which throws `produced empty Three.js bounds`. The sibling `VGroup.getCenter()` guards this with `if (this.isEmpty()) return [position]` — `Group.getCenter()` should use the same guard (the new `isEmpty()` override added just above is exactly what's needed).

src/mobjects/geometry/Line.ts
```diff
@@ -74,14 +75,14 @@ export class Line extends VMobject {
...
+   * @post result === worldMatrix * _points3D[0]
    */
   getStart(): Vector3Tuple {
-    if (this._points3D.length >= 4) {
-      const p = this._points3D[0];
-      return [p[0], p[1], p[2]];
-    }
-    return [...this._start];
+    const localPt = this._points3D.length >= 4 ? this._points3D[0] : this._start;
+    const worldMatrix = this._computeWorldMatrix();
```
maloyan, line 83: **[MAJOR] `getStart()`/`getEnd()` are now world-space, but the other accessors stayed local.** `getStart`/`getEnd` apply `_computeWorldMatrix()`, while `getMidpoint` (132), `getDirection` (142), `getLength` (121), `getAngle` (157) and `pointAlongPath` (164) still read the local `_start`/`_end`. For any rotated/scaled/shifted/parented Line these become mutually inconsistent: `(getStart()+getEnd())/2 !== getMidpoint()`, `getDirection()` no longer points from start to end, and `pointAlongPath(0) !== getStart()`. Pick one coordinate frame for all of them. (CI stays green only because no test transforms a Line before reading these.)

src/mobjects/svg/Brace.ts
```diff
@@ -80,7 +80,7 @@ export interface BraceLabelOptions extends BraceOptions {
 function getMobjectKeyPoints(mobject: Mobject): number[][] {
   // If it's a VMobject with accessible points, use those
   if (mobject instanceof VMobject) {
-    const pts = (mobject as VMobject).getPoints();
+    const pts = (mobject as VMobject).getLocalPoints();
```
maloyan: **[MAJOR] Mixing local key-points with a world-space center.** `getMobjectKeyPoints` was switched to `getLocalPoints()` (raw `_points3D`, no parent-chain transform), but `_generateBracePoints` projects each point relative to `this.mobject.getCenter()` (world-space) via `dx = p[0] - mobjectCenter[0]`, then places `braceStart`/`braceEnd` in world space. For any target that's been shifted/scaled/`moveTo`'d (or nested under a transformed group) this subtraction mixes two frames, so the brace is mis-sized and mis-positioned. On `main`, `getPoints()` was world-space and matched `getCenter()`. Use a world-space accessor here to match `getCenter()` — note the non-VMobject fallback branch already uses world-space `getCenter()`/`getBoundingBox()`, so the two branches are also inconsistent with each other right now.

rambip:
@maloyan almost done fixing. Can you manually checkout the branch and check if the examples look right ?

rambip:
I checked of all of your remarks.

PR #427, opened 2026-06-01
Normalize transform: follow-up on `copy` implementation

rambip:
## Summary

That was the missing bit of the `NormalizeTransform` refactor. I wanted to add the changes there, but bad timing.

The main thing was rewriting all of the `copy` implementation that used 2 different conventions. Now it's a lot more robust. And I catched a few other bugs regarding computing width / height of shapes.

I also added `getAllPoints`, this was useful in order to correct the implementation of `getBoundingBox`.

Remaining question: "Should we keep getLocalPoints since we don't know if it returns the object own points or all the children points too ?"

## Changes
-a lot, sorry

## Testing
- [x] Tests pass (`npm test`)

## Notes

This is probably one of the last big refactor I'm doing, after this one I will actually ship a few features (manim-web skill, fix arrow implementation).


rambip:
@maloyan ready

Still a big change, but for now I have the motivation to fix core issues and rewrite it in a clean way, so I go on.

rambip:
~And do you have any idea why I get an 'out of memory' error ?~

I fixed it

src/core/VGroup.ts
```diff
@@ -425,12 +411,11 @@ export class VGroup extends VMobject {
...
   }
 
-  /**
-   * Create a copy of this VGroup.
-   */
-  protected override _createCopy(): VMobject {
-    // Create an empty group; children are copied in Mobject.copy()
-    return new VGroup();
+  override copy(): VGroup {
+    this.normalizeTransform();
```
maloyan: **Critical: `copy()` mutates the source.** `this.normalizeTransform()` runs `_propagateTransformDownward()`, which bakes geometry and resets `this.rotation`→(0,0,0), `this.scaleVector`→(1,1,1), and rewrites positions on `this` *and every descendant*. That violates the `@post this is unchanged` contract on `_copyBaseAttributesInto` (Mobject.ts:529) — copying must never mutate the source.

Reproduced: a rotated+scaled VGroup has its source `rotation.z` go 0.785→0 and `scaleVector.x` 2→1 after `copy()`. Callers that snapshot a source mid-animation (`Transform`, `become`) get a corrupted source.

Drop the normalize on the source and let `_copyBaseAttributesInto` carry rotation/scale onto the clone (the base/non-group path already does this), or normalize the clone after construction instead.

src/core/Group.ts
```diff
@@ -182,9 +197,11 @@ export class Group extends Mobject {
   /**
    * Create a copy of this Group.
    */
-  protected override _createCopy(): Group {
-    // Create an empty group; children are copied in Mobject.copy()
-    return new Group();
+  override copy(): Group {
+    this.normalizeTransform();
```
maloyan: Same critical issue as `VGroup.copy()`: `this.normalizeTransform()` mutates the source (resets rotation/scaleVector and rewrites child geometry), breaking the `@post this is unchanged` copy contract. Normalize the clone instead of `this`, or drop it and let `_copyBaseAttributesInto` copy the transform onto the clone.

maloyan (changes requested):
Reviewed with agents + codex. The copy refactor is solid (deep-copies `_points3D` rows, re-links child parents, preserves subclass prototypes, guards `clone === this`). One **critical** issue blocks: `Group.copy()` and `VGroup.copy()` mutate the source object via `normalizeTransform()`, violating the `@post this is unchanged` contract on `_copyBaseAttributesInto` and the AGENTS.md copy doctrine. Details inline.

Verified empirically — copying a rotated+scaled VGroup resets the *source's* `rotation.z` 0.785→0 and `scaleVector.x` 2→1. World geometry is preserved (normalizeTransform keeps points), but the source's transform attributes are silently reset, which corrupts any caller that copies a source mid-animation (`Transform`, `become`, FadeTransforms/ShapeMorphStrategy all do `source.copy()`).

Fix: don't normalize the source. Either drop the `normalizeTransform()` call and let `_copyBaseAttributesInto` copy rotation/scale onto the clone (as the non-group path already does), or normalize the *clone* after building it.

Non-blocking notes (not requiring changes): a few migrated test doubles in `transform.test.ts` (`SimpleMobject`, `TextLikeMobject`) override `copy()` without calling `_copyBaseAttributesInto`, so they no longer copy base attributes — inconsistent with the production contract under test. And `getAllPoints`/`getBounds` only collect points from non-VMobject children at depth 1, so a non-VMobject (Sphere/Cube) nested deeper is omitted from bounds — edge case, fine to defer. (Codex flags on getCenter empty-fallback and getWidth=0 for stroke VMobjects were checked and are non-issues — getCenter fallback is unchanged from main, getWidth returns correct values.)

rambip:
not ready to merge right now

rambip:
@maloyan ready to re-review and merge now !

I'm finally happy with the current state, a lot simpler, a lot cleaner, and hopefully it will make it easier to catch errors in the future.

maloyan (approved):

Issue #244, opened 2026-04-12
[Feature] Json scene storage format


rambip:
## Description
Write and support a standard json format for storing scenes in order to play them.
It would be very similar to the idea of the [Vega](https://vega.github.io/vega/docs/specification/) specification for charts and interactive visualizations.

## Use Case
Having this feature would allow the user to write their scene animation in the language of their choice (like python), compile it to json, and play it to js.
In particular for python, it would enable something really cool and that a lot of user would love: a Manim widget.
I started it as a project, it's already working quite well.
You can see examples here: https://molab.marimo.io/notebooks/nb_t9fi4zkgMty1m2EuJsrgMP

It may also allow people to create tooling around it, such as a scene editor (although it's not the primary goal). 

## Proposed Solution
I have started to design a json specification file:
https://github.com/rambip/manim-widget/blob/main/spec.json

I would really appreciate your feedback on it.
The one thing I know is missing is how to handle the camera. Since it's a hard subject I'm still thinking about it.

The idea would be to implement a `player` that would play this file. My current `player.js` is less than 400 lines.
The hard part is really to have a spec that is:
- compatible with python
- future proof
- as much feature-complete as possible

rambip:
I can totally consider contributing, this issue is just a starting point for the discussion

Issue #317, opened 2026-05-04
`scene.add` should only be rendered in next `play` call


rambip:
## Description
Currently, if the user adds an object to the scene and then play an animation such as create or write, the object first appears during one frame and then is animating. That should not happen.

## Steps to Reproduce
1. `scene.add(square)`
2. `scene.play(Create(square))`

## Expected Behavior
The previous snippet should do exactly the same thing as a single scene that play create square.

## Actual Behavior
The square is added the first time, rendered during a frame, and then the animation plays.

## Environment
- Browser:
- OS:
- manim-web version:

## Additional Context
`add` should do nothing and just populate a sort of staging area for objects, then when all the animation are registered, we add only those who do not appear in animations.
No frame is rendered after `add`, all the frames are rendered during `play`

rambip:
Here is how manim does it:

====================

No frame is rendered after `scene.add()`. The first frame rendered will be the first frame of the `FadeIn` animation with opacity 0.

## The rendering flow

### No rendering on add()
`scene.add()` only modifies internal data structures - it adds the mobject to the scene's `mobjects` list but doesn't trigger any rendering [1](#3-0) .

### Rendering only happens during play()
Frames are only rendered when `scene.play()` is called. The render loop is initiated in `CairoRenderer.play()` [2](#3-1) .

### Animation setup before first frame
When `play(FadeIn(object))` is called:
1. `begin_animations()` is called, which sets up the animation's starting state [3](#3-2) 
2. `FadeIn` creates its starting mobject with opacity 0 [4](#3-3) 
3. Only then does the render loop start, beginning with this opacity 0 state

## The sequence

```python
self.add(square)      # No frame rendered - just internal state change
self.play(FadeIn(square))  # First frame rendered here with opacity 0
```

The object never appears at full opacity before the fade-in begins. The transition is smooth from opacity 0 to opacity 1 during the animation.

## Notes

- This applies to all animations - frames are only rendered during `play()` calls
- The "pop" effect mentioned earlier happens because the object's internal state changes from visible to invisible between the `add()` call and the first frame of `FadeIn`, but no intermediate frame is rendered
- If you want the object to be visible before fading, use a different approach like `self.add(square); self.wait(); self.play(FadeOut(square))`

Wiki pages you might want to explore:
- [Core Architecture (ManimCommunity/manim)](/wiki/ManimCommunity/manim#2)

### Citations

**File:** manim/scene/scene.py (L479-521)
```python
[...]

PR #447, opened 2026-06-08
Feat/play with timestamps

rambip:
## Summary

Add support for individual timings in animations.
This PR adds an experimental feature under the `scene.playWithTimestamps` method.

The main motivation for this is to support the future #244 where you can set individual timings to stored animations.

## Changes
- src/core/Scene.ts

## Testing
- [x] Tests pass (`npm test`)

## Related Issues
Support #244

### DIFF
--- a/src/core/Scene.ts
+++ b/src/core/Scene.ts
@@ -655,6 +655,79 @@ export class Scene {
   }
 
   /**
+   * Play animations with explicit per-animation start/end times (seconds).
+   *
+   * NOTE: this is a manim-widget wire-format feature — Python manim has no
+   * equivalent per-animation timing API. It is used by the player when
+   * deserializing AnimationGroup descriptors that carry `start`/`end` fields.
+   *
+   * Each entry's `end - start` overrides the animation's own duration so the
+   * Timeline schedules it for the correct window.
+   */
+  async playWithTimestamps(
+    entries: { animation: Animation; start: number; end: number }[],
+  ): Promise<void> {
+    if (entries.length === 0) return;
+
+    const animations = entries.map((e) => e.animation);
+    const allAnimations = this._collectAllAnimations(animations);
+
+    this._cancelPendingRender();
+    this._suppressAutoRender = true;
+    try {
+      const renderPromises: Promise<void>[] = [];
+      const collectRenders = (mobject: Mobject) => {
+        const asyncMob = mobject as Mobject & { waitForRender?: () => Promise<void> };
+        if (typeof asyncMob.waitForRender === 'function') {
+          renderPromises.push(asyncMob.waitForRender().catch(() => {}));
+        }
+        for (const child of mobject.children) collectRenders(child);
+      };
+      for (const anim of allAnimations) collectRenders(anim.mobject);
+      if (renderPromises.length > 0) await Promise.all(renderPromises);
+
+      for (const anim of allAnimations) {
+        if (anim.mobject._dirty) {
+          anim.mobject._syncToThree();
+          anim.mobject._dirty = false;
+        }
+      }
+
+      for (const anim of animations) anim.begin();
+
+      for (const anim of allAnimations) {
+        if (!this._mobjects.has(anim.mobject)) this.add(anim.mobject);
+      }
+
+      this._timeline = new Timeline();
+      for (const { animation, start, end } of entries) {
+        // Override duration so Timeline.add computes the correct endTime.
+        Object.defineProperty(animation, 'duration', {
+          value: end - start,
+          writable: true,
+          configurable: true,
+        });
+        this._timeline.add(animation, start);
+      }
+
+      this._timeline.play();
+      this._isPlaying = true;
+      this._currentTime = 0;
+      this._startRenderLoop();
+    } finally {
+      this._suppressAutoRender = false;
+      this._cancelPendingRender();
+    }
+
+    if (this._disposed) return;
+    await new Promise<void>((resolve) => {
+      this._playPromiseResolve = resolve;
+    });
+
+    for (const anim of allAnimations) anim.cleanUpFromScene(this);
+  }
+
+  /**
    * Play multiple animations in parallel (all at once).
    * Alias for play() - delegates to play() to avoid duplicated logic.
    * @param animations - Animations to play simultaneously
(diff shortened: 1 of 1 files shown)


src/core/Scene.ts
```diff
@@ -654,6 +654,79 @@ export class Scene {
...
+      for (const anim of animations) anim.begin();
+
+      for (const anim of allAnimations) {
+        if (!this._mobjects.has(anim.mobject)) this.add(anim.mobject);
+      }
+
+      this._timeline = new Timeline();
+      for (const { animation, start, end } of entries) {
+        // Override duration so Timeline.add computes the correct endTime.
+        Object.defineProperty(animation, 'duration', {
```
maloyan: **[major] Overriding `duration` is inert for an `AnimationGroup` — i.e. the stated primary use case.** It only time-scales *leaf* animations. For a group, `begin()` (line 696, before this loop) has already precomputed the children's `_startTimes`/`_endTimes` from their *original* durations. Then `AnimationGroup.interpolate(alpha)` does `groupTime = alpha * this.duration`, and the timeline drives `alpha = (t - start) / (end - start)`, so `groupTime = (t - start)` — the override **cancels out**. Children always run at their natural absolute timing, merely clipped/padded to the window:
- window **shorter** than the group's natural duration → later children freeze mid-animation, and `AnimationGroup.finish()` marks them finished *without* forcing `interpolate(1)` → wrong final state;
- window **longer** → children finish early, then hold.

Confirmed with a runtime probe: a sequential group of two 1s children replayed into window `[0,1]` ends with child A at α=1 but child B at α=0 (never started). To actually retime a group, rescale its child `_startTimes`/`_endTimes` by `(end-start)/computeGroupDuration(...)`, or set the group's duration *before* `begin()` and derive child times from it.

src/core/Scene.ts
```diff
@@ -654,6 +654,79 @@ export class Scene {
...
+      }
+
+      for (const anim of animations) anim.begin();
+
+      for (const anim of allAnimations) {
+        if (!this._mobjects.has(anim.mobject)) this.add(anim.mobject);
+      }
+
+      this._timeline = new Timeline();
+      for (const { animation, start, end } of entries) {
```
maloyan: **[major] No validation of `start`/`end` → the awaited play promise can hang forever.** A `NaN` (or `Infinity`) in any `start`/`end` poisons `Timeline._duration` (`endTime = startTime + duration`; `_duration = Math.max(_, NaN) = NaN`). Then `Timeline.isFinished()` is `_currentTime >= NaN` → always `false`, so `_checkFinished()` never resolves `_playPromiseResolve` and never stops the loop: the `await` at line 723 hangs and the rAF render loop spins forever (CPU/GPU). Lesser variants: `end < start` → negative duration, entry silently dropped (never interpolated); `start < 0` → silently clamped to 0 by `Timeline._resolvePosition` while the duration is still computed from the un-clamped value. Suggest validating each entry up front (finite, `end > start`, `start >= 0`) and throwing or skipping with a warning.

src/core/Scene.ts
```diff
@@ -654,6 +654,79 @@ export class Scene {
...
+          writable: true,
+          configurable: true,
+        });
+        this._timeline.add(animation, start);
+      }
+
+      this._timeline.play();
+      this._isPlaying = true;
+      this._currentTime = 0;
+      this._startRenderLoop();
```
maloyan: **[minor] Audio never starts here (diverges from `play()`).** `play()` runs `this._audioManager.seek(0); this._audioManager.play()` right before `_startRenderLoop()` (Scene.ts:629-633); this method omits it, so a scene with loaded audio replays silently. If intentional (the widget player drives audio externally), add a one-line comment saying so; otherwise restore the block — seeking to the first entry's `start` rather than 0, since timestamps are explicit.

src/core/Scene.ts
```diff
@@ -654,6 +654,79 @@ export class Scene {
...
+   * Play animations with explicit per-animation start/end times (seconds).
+   *
+   * NOTE: this is a manim-widget wire-format feature — Python manim has no
+   * equivalent per-animation timing API. It is used by the player when
+   * deserializing AnimationGroup descriptors that carry `start`/`end` fields.
+   *
+   * Each entry's `end - start` overrides the animation's own duration so the
+   * Timeline schedules it for the correct window.
+   */
+  async playWithTimestamps(
```
maloyan, line 682: **[minor] ~55 of these ~73 lines are copy-pasted from `play()`.** The render-wait, the issue #317 render-suppression + its `finally`, the dirty `_syncToThree` loop, the top-level `begin()`, add-after-begin, the `_disposed`+promise await, and `cleanUpFromScene` are all duplicated verbatim — and the comments documenting those invariants (issue #317, `waitForRender` ordering) were dropped in the copy. The only real differences are scheduling (`addParallel` vs per-entry `add` + duration override) and the omitted audio. Future fixes to `play()` won't propagate here, and this contradicts the `playAll()` note just below ("Alias for play() … to avoid duplicated logic"). Consider extracting a shared private helper, e.g. `_runScheduled(animations, scheduleOnTimeline, { startAudio })`, with both methods supplying only the scheduling callback.

src/core/Scene.ts
```diff
@@ -654,6 +654,79 @@ export class Scene {
     }
   }
 
+  /**
```
maloyan, line 670: **[minor] Document the side effect + add tests.** `Object.defineProperty(animation, 'duration', …)` permanently mutates the caller-owned `Animation` (a `readonly` field). If that same instance is later passed to `play()` it silently keeps the timestamp-derived duration; passing the same instance in two entries here corrupts both windows (the live `duration` is re-read every frame, last write wins). Worth a note in this JSDoc. Also, `Scene` is exported public API and this method ships with no tests (the PR touches only Scene.ts) despite a non-trivial timing contract — please add coverage, at least the leaf happy-path and the AnimationGroup case above. (The `defineProperty` idiom itself is fine — it matches `CameraFrame`/`AnimateProxy`.)

maloyan (commented):
Reviewed with a multi-agent pass + codex (gpt-5.5, xhigh); the findings below are the ones that survived adversarial verification (a runtime probe confirmed the AnimationGroup one). Build is green — typecheck, lint and tests all pass — and there are no call sites yet, so nothing in `main` breaks. But there are two **major** correctness issues on the feature's *primary* path, plus a few minors. Not blocking the build, but I'd address the two majors (especially the AnimationGroup one, since that's the stated use case) before this gets wired up. Details inline.

